PR #4, opened 2017-03-11
Fix completion response message

adamwalker:


alanz:
Thanks.

The JSONencodings are horrible, I am sure it must be possible to change the underlying field names to not have those horrendously long names, perhaps using overloaded record fields.

alanz:
@adamwalker I have added you as a collaborator

adamwalker:
Thanks

PR #5, opened 2017-03-11
Remove static parameter to handlers

adamwalker:
I removed the resData field in the LanguageContextData struct, as well as the parameter to each handler function that receives this data. 
As far as I can tell, this parameter is not useful because all handlers which take a static context argument can just be partially applied to it before being put in the Handlers struct.

### DIFF
--- a/src/HieVscode.hs
+++ b/src/HieVscode.hs
@@ -37,7 +37,7 @@ run :: IO Int
 run = flip E.catches handlers $ do
 
   flip E.finally finalProc $ do
-    CTRL.run (return ()) () hieHandlers hieOptions
+    CTRL.run (return ()) hieHandlers hieOptions
 
   where
     handlers = [ E.Handler ioExcept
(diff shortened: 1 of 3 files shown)


alanz:
I will want to check how that works in https://github.com/alanz/haskell-ide-engine/tree/lsp, for the rename operation, before merging

alanz:
How would you change https://github.com/alanz/haskell-ide-engine/blob/lsp/src/Haskell/Ide/Engine/Transport/LspStdio.hs#L320 to work without this?

It basically captures controller level common state, which is then provided to each handler.

In this case, it is the channel to forward requests to so that they can be processed in a separate thread.

adamwalker:
Like this: 
https://github.com/adamwalker/haskell-ide-engine/commit/d13dbb8bd3a0099fdfe4b4371e45d0de1b0fb40e
Admittedly, I haven't run it, but it builds. It should be equivalent to what you had before.

PR #6, opened 2017-03-12
Leave logger initialization to the library user

adamwalker:
The user may wish to setup logging differently to how this library currently does it. This PR removes the setupLogger call in the run function. The logging setup function is still exported for the library user to call if they wish.

### DIFF
--- a/src/Language/Haskell/LSP/Control.hs
+++ b/src/Language/Haskell/LSP/Control.hs
@@ -35,8 +35,6 @@ run :: forall a.
     -> IO Int         -- exit code
 run dp a h o = do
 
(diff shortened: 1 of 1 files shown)


alanz:
@adamwalker btw, are you building something with this library?

adamwalker:
Yeah - a VHDL language server. I'm an FPGA developer and the IDEs suck. A good language server + VSCode would be a big improvement.

alanz:
Nice. If it ever becomes public, perhaps consider listing it [here](https://github.com/Microsoft/language-server-protocol/wiki/Protocol-Implementations)

PR #9, opened 2017-04-11
DidChangeTextDocumentParams parses elisp JSON

alanz:
Which uses "null" for []

### DIFF
--- a/src/Language/Haskell/LSP/TH/DataTypesJSON.hs
+++ b/src/Language/Haskell/LSP/TH/DataTypesJSON.hs
@@ -20,8 +20,8 @@ import Data.Default
 
 -- | This data type is used to host a FromJSON instance for the encoding used by
 -- elisp, where an empty list shows up as "null"
-data List a = List [a]
-            deriving (Show,Read,Eq)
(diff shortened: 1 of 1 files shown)


src/Language/Haskell/LSP/TH/DataTypesJSON.hs
```diff
@@ -18,6 +18,20 @@ import Data.Default
 
 -- ---------------------------------------------------------------------
 
+-- | This data type is used to host a FromJSON instance for the encoding used by
+-- elisp, where an empty list shows up as "null"
+data List a = List [a]
```
adamwalker: Would a newtype be better?

alanz:
It is for the xxxJSON instances. Would that work with a newtype?

If so it does not really matter either way.

adamwalker:
I believe it would work. It doesn't really matter. 

The advantage of a newtype, I believe, is that it has no runtime overhead. List a would have exactly the same representation in memory as [a]. I don't think this is true when "data" is used.

alanz:
Ok, I will change it.

PR #11, opened 2017-04-14
initialization callback can return error

adamwalker:


src/Language/Haskell/LSP/TH/DataTypesJSON.hs
```diff
@@ -120,7 +120,7 @@ data ResponseError =
   ResponseError
     { _code    :: ErrorCode
     , _message :: String
-    , _data    :: Maybe A.Object
```
alanz: Ok, I see the error of my ways

PR #12, opened 2017-04-22
Spec v3.0

alanz:
This is pretty stable now, and the spec is backward compatible with the earlier version.


adamwalker:
Nice!
I didn't realise version 3 was backwards compatible with version 2. Does this mean that using VSCode with a version 2 language client, or neovim or emacs with plugins that support version 2 only, will still work with this version?

src/Language/Haskell/LSP/Control.hs
```diff
@@ -8,40 +8,54 @@
 module Language.Haskell.LSP.Control
   (
     run
-  , sendNotificationMessage
-  , sendRequestMessage
-  , sendResponseMessage
+  -- , sendNotificationMessage
```
alanz: In hie I store the `sendFunc` returned as the first parameter of a `Handler`, and use that.
Perhaps we should extend the `Control.run` callback function to be

```haskell
(J.ClientCapabilities -> (BSL.ByteString -> IO ()) -> IO (Maybe J.ResponseError))
```

so that the second parameter is the send function?

adamwalker: Sounds like a good idea

PR #15, opened 2017-04-30
Virtual File System

alanz:
Introduces a virtual file system, in the sense that it processes the `textDocument/didOpen`, `textDocument/didChange` and `textDocument/didClose` notifications to replicate the state of a file being edited in the IDE.

This is accesed in the `Handler` via a new field with signature

```haskell
(J.Uri -> IO (Maybe VirtualFile))
```

where a `VirtualFile` is defined as

```haskell
data VirtualFile =
  VirtualFile {
      _version :: Int
    , _text    :: Yi.YiString
    } deriving (Show)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /tags
 /cabal.project.local
 .stack-work
+/.dir-locals.el
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,19 @@
+dependencies:
+  cache_directories:
+    - "~/.stack"
+    - "~/.cabal"
+  pre:
+    - wget -q -O- https://s3.amazonaws.com/download.fpcomplete.com/ubuntu/fpco.key | sudo apt-key add -
+    - echo 'deb http://download.fpcomplete.com/ubuntu/precise stable main'|sudo tee /etc/apt/sources.list.d/fpco.list
+    - sudo apt-get update && sudo apt-get install stack -y
+  override:
+    - stack setup
+    - stack build -j 2 :
+        timeout: 3600
+    - stack build -j 2 --test --only-dependencies :
+        timeout: 3600
+
+test:
+  override:
+    - stack test haskell-lsp :
+        timeout: 3600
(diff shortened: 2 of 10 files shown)


adamwalker:
One problem with depending on the yi-rope package is that it is GPL. IANAL, so I'm not sure if that means that this package must be GPL too. This: http://stackoverflow.com/questions/11819831/including-gpl-external-library-in-mit-code seems to suggest it does.

adamwalker:
My personal, and entirely selfish preference, is to keep this library MIT so that I can use it in the VHDL language server I am building and sell it in the future.

alanz:
Yes, I was wondering about that too.

I discussed it on IRC with some people, who seemed to think that BSD3 and GPL are compatible. I am not personally worried about licensing, so would gladly change the license here to match something.

@adamwalker what are your views on that? Given you are a contributor.

alanz:
And BSD3? Because I am quite keen to have changes managed here, and YiRope is good for that.

I imagine it should be possible to pass the types/functions in as part of the initialization, if all else fails.

Except the change management is particularly easy with YiRope.

adamwalker:
Yeah - anything that uses GPL code must be GPL. It also might be possible to not have the rope stuff as part of the framework but instead provide a separate library that the user calls from within the file opened/changed/saved callbacks that does all the work of keeping track of the contents of URIs. 

But, thats not as nice as what you currently have. It's up to you really. Just consider that using yi-rope prevents commercial uses of this library.

alanz:
Well, we have identified a workaround that could be made to enable commercial use and strip out the GPL part.

So I suggest we leave it as now and let a future commercial user do that. Unless you want to do a PR on it for that functionality now already.

It is no big deal for me, I am not personally planning commercial use, but I am not keen to do much work for the benefit of commercial users only :).

If it would help, I am happy to change the license too, if you have a preference.

wiz:
> Just consider that using yi-rope prevents commercial uses of this library.

And that means any company that wants to ship a language server will have to ignore this package and build their own from scratch.

alanz:
And from GHC 8.2 we can make the Yi.Rope package a backpack specification, so you can plug in anything that meets the signature.

alanz:
@adamwalker Are you happy with the proposed workaround, if commercial use is required? I would like to merge this.

adamwalker:
Yep. The backpack proposal sounds good.

alanz:
And I captured the required cleanup in an issue, for any future commercial users.

See https://github.com/alanz/haskell-lsp/issues/16

PR #17, opened 2017-05-02
Publish diagnostics and simplify Handler interface

alanz:
It is not clear what the expected client behaviour is when receiving diagnostics from multiple sources for the same file, and when it should update what portion of the diagnostics. 

This moves management of the diagnostics into this library, where they are aggregated by document version and source, and sent as a whole when any change for a given version. If diagnostics for a new version are seen the old ones are discarded.

It also introduces the `LspFuncs` structure as the first parameter to a `Handler`, as the number of callback features is growing.

This is currently defined as

```haskell
-- | A function to send a message to the client
type SendFunc = (BSL.ByteString -> IO ())

-- | A function to publish diagnostics. It aggregates all diagnostics pertaining
-- to a particular version of a document, by source, and sends a
-- 'textDocument/publishDiagnostics' notification with the total whenever it is
-- updated.
type PublishDiagnosticsFunc = J.Uri -> Maybe J.TextDocumentVersion -> [J.Diagnostic] -> IO ()

-- | Returned to the server on startup, providing ways to interact with the client.
data LspFuncs =
  LspFuncs
    { clientCapabilities     :: !C.ClientCapabilities
    , sendFunc               :: !SendFunc
    , getVirtualFileFunc     :: !(J.Uri -> IO (Maybe VirtualFile))
    , publishDiagnosticsFunc :: !PublishDiagnosticsFunc
    }
```

### DIFF
--- a/example/Main.hs
+++ b/example/Main.hs
@@ -54,8 +54,11 @@ run dispatcherProc = flip E.catches handlers $ do
 
   let
     dp lf = do
+      liftIO $ U.logs $ "main.run:dp entered"
       atomically $ writeTChan rin (InitializeCallBack lf)
+      liftIO $ U.logs $ "main.run:dp tchan"
       dispatcherProc
+      liftIO $ U.logs $ "main.run:dp after dispatcherProc"
       return Nothing
 
   flip E.finally finalProc $ do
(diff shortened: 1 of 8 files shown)


alanz:
Forgot the link. See https://github.com/Microsoft/language-server-protocol/issues/228

PR #19, opened 2017-05-11
Simpler user API

alanz:
Introduce a new `Messages` module that mirrors the sequence in the [spec](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#messages-overview), and provides smart constructors for the messages.

This means the mapping of message methods to types happens in one place only.


alanz:
Closed via #22

PR #34, opened 2017-06-15
Clean up handler type

wz1000:


example/Main.hs
```diff
@@ -142,26 +129,21 @@ nextLspReqId = do
...
+reactor :: Core.LspFuncs -> ReactorState -> TChan ReactorInput -> IO ()
+reactor lf st inp = do
   liftIO $ U.logs $ "reactor:entered"
-  flip evalStateT st $ forever $ do
+  flip evalStateT st $ flip runReaderT lf $ forever $ do
     inval <- liftIO $ atomically $ readTChan inp
     case inval of
-      -- This will be the first message received from the client, and can be
-      -- used to tailor processing according to the given client capabilities.
-      InitializeCallBack lf@(Core.LspFuncs capabilities _sf _vf _dpf) -> do
```
alanz: Does the server get a handler call with this information, so that it can set its internal processing according to the client capablities, and advertise its own?

wz1000: That can be handled in reactor, just not in the message loop.

PR #38, opened 2017-07-02
fix haddocks error

wz1000:


alanz:
We need to add Haddock to the CI build.

PR #39, opened 2017-07-03
added plain text MarkedStrings

wz1000:


src/Language/Haskell/LSP/TH/DataTypesJSON.hs
```diff
@@ -2848,16 +2848,16 @@ Registration Options: TextDocumentRegistrationOptions
 
 -}
 
-data MarkedString =
-  -- TODO: Add the plain string variant too
-  MarkedString
+data LanguageString =
```
alanz: The spec calls it `MarkedString`. Why do you want to change the name?

wz1000: In the protocol,

    type MarkedString = string | { language: string; value: string };

I simply named { language: string; value: string } as LanguageString 

There is still a MarkedString in DatatypesJSON.

    type MarkedString = Either Text LanguageString

PR #50, opened 2017-08-01
Fix inconsistent NotificationParams naming scheme

### LINKED ISSUE #49: Inconsistency in naming of NotificationParams
The notification params for the `Initialized` notification are called `InitializedParams`, but the params for the `Exit` notification are called `ExitNotificationParams`. I think they should either both contain the word `Notification` or neither.

I personally think `ExitNotificationParams` should be changed to `ExitParams`. `TraceNotificationParams` should then also be renamed to `TraceParams`.

noughtmare:
Fixes #49


wz1000:
Can you enable CircleCI for your fork?

PR #63, opened 2017-12-04
Don't escape semicolons after drive letters

nponeccop:
The RFC says that semicolons may be used for special purposes in path parts of urls, and in this case no escaping is allowed. And this is consistent with Microsoft samples.

See https://github.com/haskell/haskell-ide-engine/issues/329 for discussion
See https://github.com/autozimu/LanguageClient-neovim/issues/144 for normative references and Microsoft examples of typical errors in URLs

### DIFF
--- a/src/Language/Haskell/LSP/TH/DataTypesJSON.hs
+++ b/src/Language/Haskell/LSP/TH/DataTypesJSON.hs
@@ -62,7 +62,7 @@ uriToFilePath (Uri uri)
 
 filePathToUri :: FilePath -> Uri
 filePathToUri (drive:':':rest) =
-  Uri $ T.pack $ concat ["file:///", [toLower drive], "%3A", fmap convertDelim rest]
+  Uri $ T.pack $ concat ["file:///", [toUpper drive], ":", fmap convertDelim rest]
   where
     convertDelim '\\' = '/'
     convertDelim c = c


nponeccop:
@alanz ping :)

alanz:
@wz1000 is currently reworking things, that includes bringing in a different filepath in ghc-mod. Once that is done we can confirm whether this issue still exists.

alanz:
And raeading more closely, this is not the issue I thought it was :(

nponeccop:
So what are we going to do? Correct per specification url escaping is a good thing as we deal with them all the time.

alanz:
We will fix it, I read the backscroll and got confused about what precisely it was.

But I first want the @wz1000 async stuff to land, as it touches a lot of the repos involved in this one, and I don't wont a tricky merge again.

The best will be to use an actual URI type, so that escaping gets handled as per the spec.

nponeccop:
Oh, I see. We all hate tricky merges :)

nponeccop:
@alanz Can we merge it now?

@soylenz at https://github.com/haskell/haskell-ide-engine/issues/396 seems to report that `%3a` still happens after the @wz1000 merge.

soylens:
Have you not been experiencing the `%3a` since @wz1000 's commit?

nponeccop:
I haven't tested hie on windows after the commit yet.

PR #66, opened 2017-12-13
Add Foldable and Traversable instance to List (fixes #65)

### LINKED ISSUE #65: Make List Traversable
That would allow me to do something like this:

```haskell
let m = fromJust (decode message) :: LSP.PublishDiagnosticsNotification
mapM_ T.putStrLn $ m ^.. LSP.params . LSP.diagnostics . traverse . LSP.message
```

noughtmare:


alanz:
For some reason I don't get notification about PRs. So perhaps ping me when creating one.

Issue #16, opened 2017-05-02
Commercial user GPL cleanup


alanz:
https://github.com/alanz/haskell-lsp/pull/15 introduced a GPL library for the Virtual File System Feature.

This is a performance optimisation, and is localised to that module.

Any future commercial user will have to file a PR to either

- replace the library with an alternative (it is basically a `Data.Text` interface)
- let the initial setup include a structure containing the methods from `Yi.Rope` that are used, and pass in a non-GPL variant for commercial use, thus preserving performance for non-commercial users
- From GHC 8.2, use backpack to define a mixin with the features required, and use some non-GPL alternative

This issue simply captures the state of play, I (@alanz) have no intention of fixing it.

cocreature:
Given that the PR to switch to rope-utf16-splay has been merged, I believe that this issue can be closed.

PR #70, opened 2018-02-10
Switch to using the rope-utf16-splay library for ropes

ollef:
* This makes the licensing story simpler because rope-utf16-splay is
BSD3 licensed. (#16)
* LSP uses UTF-16 code unit based indexing, while the Yi rope library
indexes by characters. This means that haskell-lsp would previously not
correctly count characters that are encoded as two code units in
UTF-16. The rope-utf16-splay library uses code unit indexing, which
fixes this issue.
* From my benchmarks, this new library, which uses splay trees
internally, can be about twice as fast as finger tree based ones (like
the Yi one) for use cases that are similar to what haskell-lsp might be
doing (many consecutive modifications).

### DIFF
--- a/test/VspSpec.hs
+++ b/test/VspSpec.hs
@@ -2,9 +2,10 @@
 module VspSpec where
 
 
+import           Data.String
+import qualified Data.Rope.UTF16 as Rope
 import           Language.Haskell.LSP.VFS
 import qualified Language.Haskell.LSP.Types as J
-import qualified Yi.Rope as Yi
 
 import           Test.Hspec
 
@@ -62,9 +63,9 @@ vspSpec = do
           , "-- fooo"
           , "foo :: Int"
           ]
-        new = applyChange (Yi.fromString orig)
+        new = applyChange (fromString orig)
                 $ J.TextDocumentContentChangeEvent (mkRange 2 1 2 5) (Just 4) ""
-      lines (Yi.toString new) `shouldBe`
+      lines (Rope.toString new) `shouldBe`
           [ "abcdg"
           , "module Foo where"
           , "-oo"
@@ -79,9 +80,9 @@ vspSpec = do
           , "-- fooo"
           , "foo :: Int"
           ]
-        new = applyChange (Yi.fromString orig)
+        new = applyChange (fromString orig)
                 $ J.TextDocumentContentChangeEvent (mkRange 2 1 2 5) Nothing ""
-      lines (Yi.toString new) `shouldBe`
+      lines (Rope.toString new) `shouldBe`
           [ "abcdg"
           , "module Foo where"
           , "-oo"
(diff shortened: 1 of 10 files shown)


MaskRay:
https://github.com/Microsoft/language-server-protocol/issues/376

test/VspSpec.hs
```diff
@@ -245,3 +248,15 @@ vspSpec = do
...
           ]
+    it "indexes using utf-16 codepoints" $ do
+      let
+        orig = unlines
+          [ "a𐐀b"
+          , "a𐐀b"
+          ]
+        new = applyChange (Yi.fromString orig)
+                $ J.TextDocumentContentChangeEvent (mkRange 1 0 1 3) (Just 3) "𐐀𐐀"
+      lines (Yi.toString new) `shouldBe`
```
wz1000: Yi.toString new

Does this compile?

wz1000:
Could you enable CircleCI for your fork?

wz1000:
Also how does this compare with yi-rope on memory usage?

test/VspSpec.hs
ollef: Thanks, I must have messed up a rebase. Fixed!

ollef:
@wz1000: CI enabled.

I have yet to run any memory benchmarks but it uses the strategy of merging adjacent chunks if they're below a certain size just like Yi.Rope, so it should be very similar in terms of memory usage.

ollef:
@MaskRay: Can you explain the relevance of that link to this PR? I agree with it, but as far as I can tell that change hasn't been made yet and it seems uncertain when or if it will be.

alanz:
What is the state of play on this?

I see it needs an update, and making sure it has GHC 8.4.2 support would be great too.

In which case a merge is probably in order.

Does anyone have reservations? @wz1000 @MaskRay ?

ollef:
I kind of forgot about it. I'll get the library updated and conflicts fixed when I have time.

I'd also be interested in hearing about reservations, e.g. if any of the reverse dependencies are strongly committed to using `yi-rope`.

alanz:
I want to first sort out https://github.com/haskell/haskell-ide-engine/issues/538 before landing this, do not want to confuse possible sources of problems.

ollef:
That's fair and there's no rush from my end. Feel free to ping me when the time's right! :)

mpickering:
Seems like https://github.com/haskell/haskell-ide-engine/issues/538 has been closed now. Would be good to get this merged I think.

ollef:
I've rebased this on master now.

alanz:
Given https://github.com/Microsoft/language-server-protocol/issues/376 and https://github.com/Avi-D-coder/lsp-range-unit-survey we should perhaps hold off until we know if anything is going to change.

ndmitchell:
Could I suggest this gets merged sooner rather than later? Anyone hoping to build on haskell-lsp in a BSD-only context is blocked without it (which is me 😄 - see https://github.com/digital-asset/haskell-lsp and 15096280b3542d56d8df728d3bce9ea34141debd which is based on this patch). It seems from the links above that this is the right approach today. If it stops being the right approach that seems the right time to go off in a different direction.

alanz:
Ok, I guess we can do it, and it is early in the month, so the bleeding edger's can pick up any problems that arise.

Issue #68, opened 2017-12-15
Error with a fresh install from haskell-ide-engine master because of initialized message


tolgap:
This is the error I run into:

```
2017-12-15 13:22:40.918753 [ThreadId 4] - ---> {"jsonrpc":"2.0","method":"initialized"}
2017-12-15 13:22:40.919144 [ThreadId 5] - <--2--{"jsonrpc":"2.0","method":"window/logMessage","params":{"type":1,"message":"haskell-lsp:parse error. Object (fromList [(\"jsonrpc\",String \"2.0\"),(\"method\",String \"initialized\")]) \"When parsing the record NotificationMessage of type Language.Haskell.LSP.TH.DataTypesJSON.NotificationMessage the key params was not present.\" `stack update` and install new haskell-lsp. Or check information on https://marketplace.visualstudio.com/items?itemName=xxxxxxxxxxxxxxx"}}
```

which is correct. The initialized notification does not have a "params" key, while the [`NotificationMessage` record](https://github.com/alanz/haskell-lsp/blob/b747454c06686620762dcb6e35564dd15361c7d4/src/Language/Haskell/LSP/TH/DataTypesJSON.hs#L392) needs one.

is this somehow a version mismatch? I've tried changing my stack.yaml in haskell ide engine to the 0.2 release commit

```
- location:
    git: https://github.com/alanz/haskell-lsp.git
    commit: b747454c06686620762dcb6e35564dd15361c7d4
  extra-dep: true
```

and rebuild it, but that didn't help either.

robrix:
I’m getting a similar error on `DidChangeWatchedFilesParams`:

```
2018-03-18 17:21:17.918103 [ThreadId 4] - ---> {"jsonrpc":"2.0","method":"workspace/didChangeWatchedFiles","params":{"changes":[{"uri":"file:///Users/rob/Developer/Projects/semilattices/src/Data/Semilattice/Join.hs","type":2}]}}
[...]

PR #72, opened 2018-03-18
Correct the name of the DidChangeWatchedFilesParams field.

robrix:
This PR corrects the name of the `DidChangeWatchedFilesParams` field to match [the specification](https://microsoft.github.io/language-server-protocol/specification#workspace_didChangeWatchedFiles) (to wit, `changes` and not `params`).

While I never had the similar `initialized` error described in #68, this fixes the error messages I described in https://github.com/alanz/haskell-lsp/issues/68#issuecomment-374049116.


robrix:
Thank you!

Issue #47, opened 2017-08-01
Error decoding InitializeResponse


noughtmare:
I'm not sure if this is an issue with `haskell-lsp`, but hopefully somebody here can help me.

I have written a test LSP client, but I get an error when trying to decode the a response from `haskell-ide-engine`.

I have saved (and cleaned up and formatted) the result of an `InitializeRequest` in `result.json`, the contents are:

```json
{
    "result":{
        "capabilities":{
            "textDocumentSync":{
                "openClose":true,
                "change":2,
                "willSave":false,
                "willSaveWaitUntil":false,
                "save":{
                    "includeText":false
                }
            },
            "hoverProvider":true,
            "completionProvider":{
                "resolveProvider":true,
                "triggerCharacters":["."]
            },
            "definitionProvider":true,
            "documentHighlightProvider":true,
            "documentSymbolProvider":true,
            "codeActionProvider":true,
            "documentFormattingProvider":true,
            "documentRangeFormattingProvider":true,
            "renameProvider":true,
            "executeCommandProvider":{
                "commands":["applyrefact:applyOne","hare:demote"]
            }
        }
    },
    "jsonrpc":"2.0",
    "id":0
}
``` 

But when I try to decode it with aeson in ghci I get this:

```haskell
> eitherDecode <$> B.readFile "result.json" :: IO (Either String LSP.InitializeResponse)
Left "Error in $.result.capabilities.textDocumentSync: mempty"
```

If I just decode the `textDocumentSync` part it work correctly.

I saved this to `textDocumentSync.json`:

```json
{
    "openClose":true,
    "change":2,
    "willSave":false,
    "willSaveWaitUntil":false,
    "save":{
        "includeText":false
    }
}
```

```haskell
> eitherDecode <$> B.readFile "textDocumentSync.json" :: IO (Either String LSP.TextDocumentSyncOptions)
Right (TextDocumentSyncOptions {_openClose = Just True, _change = Just TdSyncIncremental, _willSave = Just False, _willSaveWaitUntil = Just False, _save = Just (SaveOptions {_includeText = Just False})})
```

alanz:
The client side has not really received first class support up to now.

I have been meaning to put some tests into place which runs a mock client, to drive usage scenarios, both to test the software and document usages. But time is the problem.

And it really needs some roundtrip quickcheck/hedgehog tests on the serialisation.

wz1000:
This works for me

```
[...]

PR #76, opened 2018-04-30
Make sure to escape URIs properly for Windows file paths

### LINKED ISSUE #75: URLs are not correctly encoded when publishing diagnostics
[The spec](https://microsoft.github.io/language-server-protocol/specification#basic-json-structures) says that URIs should be encoded according to RFC3986; incoming URIs are decoded in `uriToFilePath` but not encoded in `filePathToUri`.

(I am working on fixing this!)

See also: [vscode-hie-server #47](https://github.com/alanz/vscode-hie-server/issues/47)

johnsonwj:
Fixes #75. Also added a couple of pretty dumb tests!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 /cabal.project.local
 .stack-work
 /.dir-locals.el
+/.vscode
--- a/haskell-lsp.cabal
+++ b/haskell-lsp.cabal
@@ -46,6 +46,7 @@ library
                      , hashable
                      , lens >= 4.15.2
                      , mtl
+                     , network-uri
                      , parsec
                      , sorted-list == 0.2.0.*
                      , stm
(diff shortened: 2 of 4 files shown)


wz1000:
Looks good. Could you enable CircleCI for your fork?

This might also be of some interest: https://github.com/ekmett/coda/blob/master/lib/coda-lsp/Language/Server/Protocol.hs#L489

johnsonwj:
Sweet, circleci is enabled!

It looks like `coda-lsp` has the same problem...I will look into putting in a PR there too!

johnsonwj:
PS - I am trying to verify that this is in fact the root cause of [vscode-hie-server #47](https://github.com/alanz/vscode-hie-server/issues/47), but I am having trouble getting VS Code to run HIE with my local build of `haskell-lsp`. So that is pending. 

PPS - given that this filepath <--> URI issue appears in more than one place, would this be a good candidate for a hackage library or will copypasting to `coda-lsp` be sufficient?

alanz:
@johnsonwj if you want to make a library it will help the community.

But in the mean time I am going to merge this. Thanks.

PR #79, opened 2018-05-01
Split out a separate haskell-lsp-types package.

alanz:
This makes it easier/faster to work on the main package, without
running all the TH generation every compile.

cc @Bubba


noughtmare:
I think this would also allow [my haskell-lsp-client library](https://github.com/noughtmare/haskell-lsp-client) library to depend only on the types.

PR #81, opened 2018-05-03
Make relatedInformation a Maybe List to match the spec

Bubba:
The specification defines relatedInformation as optional, so this makes it a Maybe type. This fixes HIE with Visual Studio Code


Bubba:
haskell-ide-engine also needs some fixes for this API change so a new version of this might need to be published

PR #86, opened 2018-05-30
Change sendFunc to take in OutMessage instead of just any ToJSON

Bubba:
This also adds a handlers for initialisation requests and exit notifications that can be used by other packages.
We might also want to look at splitting `OutMessage` into separate types for messages from the client and messages from the server.


Bubba:
Superseeded by #87

PR #87, opened 2018-05-31
Split out OutMessage and change "recording" to "capturing"

Bubba:
`OutMessage` now consists of two separate data types for messages that originate from the client and messages that originate from the server.

Recording is now called capturing and encodes message types based on these new types. It now saves the session to one file with timestamped messages and type information.

### DIFF
--- a/src/Language/Haskell/LSP/Control.hs
+++ b/src/Language/Haskell/LSP/Control.hs
@@ -15,6 +15,7 @@ module Language.Haskell.LSP.Control
 import           Control.Concurrent
 import           Control.Concurrent.STM.TChan
 import           Control.Concurrent.STM.TVar
+import           Control.Exception
 import           Control.Monad
 import           Control.Monad.STM
 import qualified Data.Aeson as J
@@ -26,6 +27,7 @@ import           Data.Monoid
 import qualified Language.Haskell.LSP.Core as Core
 import           Language.Haskell.LSP.Utility
 import           System.IO
+import           System.IO.Error
 import           System.Directory
 import           Text.Parsec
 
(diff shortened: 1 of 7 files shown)


src/Language/Haskell/LSP/Control.hs
```diff
@@ -68,41 +68,44 @@ runWithHandles hin hout dp h o recInFp recOutFp = do
   hSetEncoding  hout utf8
 
   -- Delete existing recordings if they exist
-  forM_ [recInFp, recOutFp] $ maybe (return ()) $ \f -> do
-    exists <- doesFileExist f
-    when exists $ removeFile f
+  maybe (return ()) removeFileIfExists captureFp
```
alanz: perhaps capture against a filename with a timestamp in it.  That way we do not lose something accidentally.

Eventually, that is.

alanz:
Is there a matching change for hie?

Bubba:
Yep, in https://github.com/Bubba/haskell-ide-engine/tree/capturing

src/Language/Haskell/LSP/Control.hs
```diff
@@ -67,12 +67,12 @@ runWithHandles hin hout dp h o captureFp = do
   hSetBuffering hout NoBuffering
   hSetEncoding  hout utf8
 
-  -- Delete existing recordings if they exist
-  maybe (return ()) removeFileIfExists captureFp
-
+  timestamp <- formatTime defaultTimeLocale (iso8601DateFormat (Just "%H-%M-%S")) <$> getCurrentTime
```
alanz: Probably best to add yymmdd as well.

Bubba: [iso8601DateFormat](https://hackage.haskell.org/package/time-1.9.1/docs/Data-Time-Format.html#v:iso8601DateFormat) adds that for us

PR #90, opened 2018-06-24
Split out DataTypesJSON into smaller modules

Bubba:
Start splitting up `DataTypesJSON.hs` into smaller modules.
Needed to do this since ClientCapabilities now references CodeAction types, which then led into a whole chain of module dependencies. 
Some caveats to look out for:
- `Language.Haskell.LSP.TH.Messages` is now `Language.Haskell.LSP.TH.MessageFuncs` to make way for `Language.Haskell.LSP.TH.Message`, which contains the actual message types
- Lenses still need to be generated in `DataTypesJSON.hs`, otherwise we get duplicate functions and classes which can't be exported.


ChangeLog.md
```diff
@@ -1,5 +1,13 @@
 # Revision history for haskell-lsp
 
+## 0.4.0.0
```
alanz: I think we need to move this change to the next commit, the rest looks good.

PR #94, opened 2018-06-25
Add the root directory to LspFunc so that HIE can access it

Bubba:


alanz:
Perhaps remove the default instance for CodeActionKindValueSet from this PR? It seems to be unrelated.

Bubba:
@alanz Whoops, that was meant to be in the other branch

PR #97, opened 2018-06-30
Fix missing <> on 8.2.2 and below

Bubba:
Need to import Data.Monoid


alanz:
Does this work for later GHC's too? Without warnings?

The "usual" fix is to do

```haskell
#if __GLASGOW_HASKELL__ < 804
import           Data.Semigroup
#endif
```

alanz:
And I just tried it locally, seems to work

PR #107, opened 2018-08-01
Fix platformAwareUriToFilePath for empty paths

Anrock:
This PR makes `platformAdjustFromUriPath` act as `id` for empty paths.

Ref https://github.com/haskell/haskell-ide-engine/issues/713
Some tests in `test/unit/JsonSpec` fail when empty strings are passed to `platformAwareUriToFilePath`, then `splitDirectories` return empty list which then fails to pattern match on `firstSegment:rest`.

I'm not sure if empty string is a valid value for `srcPath` or property tests should be altered to produce non-empty strings, though.


Anrock:
@Bubba fyi

PR #126, opened 2018-08-07
Update server capabilities to 3.10

Bubba:


.circleci/config.yml
```diff
@@ -41,8 +41,8 @@ defaults: &defaults
           - ~/.stack
           - ~/.cache
 
-    - restore_cache:
-          key: stack-work-dirs-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "stack-build.txt" }}
+#     - restore_cache:
```
alanz: What is going on here? Rather restore a cache with a sequence number in it, as we do for hie. that way it fails the first time, but gets used from then onward.

Bubba: Should we move the "Compile but do not run tests" step to after it then?

alanz: This is modeled on the one for hie, and I see we have the index number in it for that. So perhaps keep them basically the same.

PR #127, opened 2018-08-07
Add client capability lenses and rename TH to Types

### LINKED ISSUE #125: Add lenses for ClientCapabilities
The main types have generated lenses, so the client capabilities should probably also have them for consistency

Bubba:
Lenses all need to be compiled in one module otherwise they get duplicate type classes which conflict.  Closes #125

### DIFF
--- a/src/Language/Haskell/LSP/Types.hs
+++ b/src/Language/Haskell/LSP/Types.hs
@@ -1,8 +1,0 @@
-module Language.Haskell.LSP.Types
-  (
-    module Language.Haskell.LSP.TH.DataTypesJSON
-  , module Language.Haskell.LSP.TH.Lens
-  ) where
-
-import Language.Haskell.LSP.TH.DataTypesJSON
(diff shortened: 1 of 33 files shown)


src/Language/Haskell/LSP/Types.hs
```diff
@@ -1,6 +1,8 @@
 module Language.Haskell.LSP.Types
   (
     module Language.Haskell.LSP.TH.DataTypesJSON
+  , module Language.Haskell.LSP.TH.Lens
```
Bubba: I was thinking we should maybe move `Language.Haskell.LSP.Types` into `haskell-lsp-types` and just reexport it from the .cabal file, rather than having this stub module in `haskell-lsp`

Would it be also worthwhile to separate Lens as well, like `Language.Haskell.LSP.Types.Lens`? I've seen a handful of packages on hackage that handle it this way but it would be a very breaking change

PR #140, opened 2019-02-27
Update Handler to delegate to typeDefinitionHandler

fendor:
It seems like a mistake? Or was that intentional?
Related to [#31](https://github.com/bubba/lsp-test/pull/31).


alanz:
Thanks for spotting this.

PR #145, opened 2019-04-10
Add some utils for TextEdit

bubba:
Originally created for HIE internals, but thought they might as well reside beside the types for others to use.


alanz:
Thanks

PR #147, opened 2019-04-11
Set the typedefinitionProvider capability if it has a handler

alanz:
The capability can either be a static bool, or a dynamic option given
the specific file it applies to.  If the server explicitly sets the
option to one of those in its Config we use it, else set it to
statically True if there is a handler for the type definition request.


alanz:
cc @fendor

Issue #144, opened 2019-04-04
Add support for window/progress


mpickering:
https://github.com/Microsoft/vscode-languageserver-node/pull/261

Vscode and sourcegraph clients implement this at least.

bubba:
Should `haskell-lsp` provide a function in `LspFuncs` to communicate progress? Or is it enough just to expose the types and let the server take care of sending the notifications?

alanz:
I think the more we can push into haskell-lsp the better.  It has a force-multiplier effect.

PR #150, opened 2019-04-20
Add window/progress notification

bubba:
Part of #144


haskell-lsp-types/src/Language/Haskell/LSP/Types/Window.hs
```diff
@@ -0,0 +1,269 @@
...
+  parseJSON (A.Number 3) = pure MtInfo
+  parseJSON (A.Number 4) = pure MtLog
+  parseJSON _            = mempty
+
+-- ---------------------------------------
+
+
+data ShowMessageParams =
+  ShowMessageParams {
+    _xtype   :: MessageType
```
alanz: I presume `_xtype` is used rather than `_type` to prevent a problem with using the generated lens being a keyword.

alanz:
If my one comment is correct, then we can merge this.

haskell-lsp-types/src/Language/Haskell/LSP/Types/Window.hs
bubba: I believe so. This was already there, just shuffled it out of `DataTypesJSON.hs`

PR #154, opened 2019-04-22
Make callback in withProgress IO

bubba:
In order for the callback to passed to monads such as Ghc, it shouldn't be constrained to be the same as the monad that it is run in


alanz:
LGTM

bubba:
cc @mpickering

PR #157, opened 2019-04-26
Improvements to window/progress/cancel

mpickering:
I'm not sure that the approach taken to cancellation is correct. It at least needs to be made configurable and the `withProgress` function needs to indicate that the progress is cancellable. 

cc @bubba


alanz:
@bubba @mpickering I will trust your guys' judgement on this one, merge when you are both happy.

bubba:
Don't merge yet, need to fix some issues with parsing the `window/progress/cancel` notification

PR #161, opened 2019-05-05
Bring over functions from @mpickering's hie-bios

alanz:


alanz:
cc @mpickering

src/Language/Haskell/LSP/Core.hs
```diff
@@ -417,6 +419,32 @@ hwf h tvarDat json = do
 getVirtualFile :: TVar (LanguageContextData c) -> J.Uri -> IO (Maybe VirtualFile)
 getVirtualFile tvarDat uri = Map.lookup uri . resVFS <$> readTVarIO tvarDat
 
+persistVirtualFile :: TVar (LanguageContextData c) -> J.Uri -> IO FilePath
```
bubba: We should probably add some haddock comments to these at some point

PR #170, opened 2019-05-27
Fix filepath to uri conversion

cocreature:
We need to escape characters in the individual path components
otherwise all that’s necessary to break the conversion is a space.
This commit also adds roundtrip tests to ensure that we don’t break
this again in the future.

### DIFF
--- a/test/URIFilePathSpec.hs
+++ b/test/URIFilePathSpec.hs
@@ -1,11 +1,17 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE OverloadedStrings #-}
 module URIFilePathSpec where
 
+import Data.List
+import Data.Monoid ((<>))
 import Data.Text                              (pack)
 import Language.Haskell.LSP.Types
 
 import           Network.URI
+import qualified System.FilePath.Windows as FPW
 import Test.Hspec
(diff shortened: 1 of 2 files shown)


test/URIFilePathSpec.hs
```diff
@@ -61,8 +65,8 @@ filePathUriSpec = do
     theFilePath `shouldBe` (Uri "file:///./Functional.hs")
 
   it "converts a Windows file path to a URI" $ do
-    let theFilePath = platformAwareFilePathToUri windowsOS "c:./Functional.hs"
-    theFilePath `shouldBe` (Uri "file:///c%3A/./Functional.hs")
+    let theFilePath = platformAwareFilePathToUri windowsOS "c:/Functional.hs"
+    theFilePath `shouldBe` (Uri "file:///c:/Functional.hs")
```
cocreature: Colons do not need to be escaped, see https://blogs.msdn.microsoft.com/ie/2006/12/06/file-uris-in-windows/

cocreature: I don’t know why the test was adding a `/` implicitely. Afaict that shouldn’t happen.

alanz:
LGTM, thanks.

PR #171, opened 2019-05-31
Add support for custom request and notification methods

cocreature:
It is relatively common to extend LSP with custom method types that
are specific to a given language so it makes sense for haskell-lsp to
support this.

To accomplish this, this PR introduces the following changes:

1. Remove the `Misc` constructor in favor of a more explicit
   `CustomClientMethod` constructor that does not enforce that methods
   start with $/. While the spec mentions that custom methods should
   start with $/, I’ve seen custom methods that do not start with $/
   and given that haskell-lsp can only control the server side but not
   the client side, e.g., VSCode extensions, I think it makes more
   sense to be relaxed here.

2. Add `ReqCustomClient` and `NotCustomClient` to `FromClientMessage`
   while removing `UnknownFromClientMessage` (any valid JSON-RPC
   message should now be decodable).

3. Add handlers for `ReqCustomClient` and `NotCustomClient`.

4. It adds `ReqCustomServer`, `RspCustomServer` and `NotCustomServer`
   to `FromServerMessage`.

Note that this PR does not add `RspCustomClient`. This is because
responses are always decoded to `BareResponseMessage` with no
restrictions so it would be redundant. Once `BareResponseMessage` is
removed in favor of explicit response types, this constructor and a
corresponding handler should be added.

### DIFF
--- a/haskell-lsp-types/src/Language/Haskell/LSP/Types/DataTypesJSON.hs
+++ b/haskell-lsp-types/src/Language/Haskell/LSP/Types/DataTypesJSON.hs
@@ -841,6 +841,14 @@ Notification:
 
 type TelemetryNotification = NotificationMessage ServerMethod A.Value
 
+type CustomClientNotification = NotificationMessage ClientMethod A.Value
+type CustomServerNotification = NotificationMessage ServerMethod A.Value
+
+type CustomClientRequest = RequestMessage ClientMethod A.Value A.Value
+type CustomServerRequest = RequestMessage ServerMethod A.Value A.Value
+
+type CustomResponse = ResponseMessage A.Value
+
 -- ---------------------------------------------------------------------
 {-
 New in 3.0
--- a/haskell-lsp-types/src/Language/Haskell/LSP/Types/Message.hs
+++ b/haskell-lsp-types/src/Language/Haskell/LSP/Types/Message.hs
@@ -8,9 +8,7 @@ import           Data.Aeson.TH
 import           Data.Aeson.Types
 import           Data.Hashable
 -- For <= 8.2.2
-import           Data.Monoid                                ((<>))
 import           Data.Text                                  (Text)
-import qualified Data.Text                                  as T
 import           Language.Haskell.LSP.Types.Constants
 
 
@@ -115,9 +113,8 @@ data ClientMethod =
  | TextDocumentOnTypeFormatting
  | TextDocumentRename
  | TextDocumentFoldingRange
- -- Messages of the form $/message
- -- Implementation Dependent, can be ignored
- | Misc Text
+ -- A custom message type. It is not enforced that this starts with $/.
+ | CustomClientMethod Text
    deriving (Eq,Ord,Read,Show)
 
 instance A.FromJSON ClientMethod where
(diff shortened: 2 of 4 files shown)


haskell-lsp-types/src/Language/Haskell/LSP/Types/DataTypesJSON.hs
```diff
@@ -841,6 +841,12 @@ Notification:
 
 type TelemetryNotification = NotificationMessage ServerMethod A.Value
 
+type CustomNotification = NotificationMessage Text A.Value
```
wz1000: Shouldn't this be
```suggestion
type CustomNotification = NotificationMessage ClientMethod A.Value
```

cocreature: That would require adding a new constructor to `ClientMethod` that handles this. I don’t think we gain anything from this. This really can be an arbitrary Text. The reason for why we need to add the corresponding constructor to `ServerMethod` is that we need to be able to match on it in `handlerMap`. This does not apply here.

wz1000: Still, it might be worth it for the sake of consistency, everything else is of the form `NotificationMessage ClientMethod` or `NotificationMessage ServerMethod`, and I believe there are functions for sending/receiving messages that constrain this type parameter to `ClientMethod` or `ServerMethod` so that you don't accidentally send/receive the wrong kind of message.

cocreature: > Isn't `CustomClientMessage` the required `ClientMethod` constructor?

Oh right, I mixed things up. I’m happy to change it to `ClientMethod` if you and @alanz prefer ithat but personally I think that’s more confusing than helpful. It suggests that it can be any of the constructors of `ClientMethod` when really we only want to allow for `CustomClientMethod`.

wz1000: > It suggests that it can be any of the constructors of ClientMethod when really we only want to allow for CustomClientMethod.

Yes, but it is like that for all of the other types too. This was a compromise as opposed to a more heavyweight system with singletons.

This way, we get one extra bit of information at the type level(whether something is a ClientMethod or a ServerMethod).

cocreature: Alright, I’ve switched to using `ClientMethod` and `ServerMethod`.

bubba:
This should close https://github.com/alanz/haskell-lsp/pull/171/

cocreature:
I pushed a minor fix to handle `CustomServerMethod` in the `FromJSON` instance. Apart from that, we have been using this since I opened this PR and both server-side and also client-side via `lsp-test` (I’ll open a PR there once this one gets merged) and it seems to work quite well.

alanz:
@wz1000 are you happy with this? In which case it can merge.

wz1000:
@alanz yes

cocreature:
Thank you!

PR #172, opened 2019-06-07
Fix decoding of ResponseMessage

cocreature:
`result = Just Nothing` gets encoded to `result = null` in
JSON. However, upon decoding that gives us `result = Nothing` so we
cannot roundtrip. This is obviously undesirable and also breaks the
assumption that exactly one of `result` or `error` must be `Just`. As
far as I know, there is know way to get this behavior by setting Aeson
options, so I switched to a hand-written FromJSON instance for
`ResponseMessage`.

### DIFF
--- a/haskell-lsp-types/src/Language/Haskell/LSP/Types/Message.hs
+++ b/haskell-lsp-types/src/Language/Haskell/LSP/Types/Message.hs
@@ -384,7 +384,16 @@ data ResponseMessage a =
     , _error   :: Maybe ResponseError
     } deriving (Read,Show,Eq)
 
-deriveJSON lspOptions ''ResponseMessage
+deriveToJSON lspOptions ''ResponseMessage
+
+instance FromJSON a => FromJSON (ResponseMessage a) where
+  parseJSON = withObject "Response" $ \o ->
+    ResponseMessage
+      <$> o .: "jsonrpc"
+      <*> o .: "id"
+      -- It is important to use .:! so that result = null gets decoded as Just Nothing
+      <*> o .:! "result"
+      <*> o .:! "error"
 
 type ErrorResponse = ResponseMessage ()
 
--- a/test/JsonSpec.hs
+++ b/test/JsonSpec.hs
@@ -31,16 +31,17 @@ jsonSpec :: Spec
 jsonSpec = do
   describe "General JSON instances round trip" $ do
   -- DataTypesJSON
-    prop "LanguageString"    (propertyJsonRoundtrip :: LanguageString -> Bool)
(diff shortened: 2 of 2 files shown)


alanz:
@bubba are you ok with this?

bubba:
Would it not be more accurate to encode `result = Just Nothing` as:
```json
{
  "result": {}
}
```

Also do there exist any types of `ResponseMessage (Maybe a)` ?

bubba:
@cocreature I do agree though that it might be nice to encode the result/error invariant as an `Either` or something, since the specification does enforce it:

```typescript
interface ResponseMessage extends Message {
	/**
	 * The request id.
	 */
	id: number | string | null;

	/**
	 * The result of a request. This member is REQUIRED on success.
	 * This member MUST NOT exist if there was an error invoking the method.
	 */
	result?: string | number | boolean | object | null;

	/**
	 * The error object in case a request fails.
	 */
	error?: ResponseError<any>;
}
```

cocreature:
> ```json
> ```json
> 
> ```
> 
> 
> Also do there exist any types of ResponseMessage (Maybe a) ?
> ```

Yes hover, this is how I noticed this bug.

> Would it not be more accurate to encode `result = Just Nothing` as:

I don’t think so, for hover at least you should encode to `result = null` according to the spec. When the spec says that the member must exist they are referring to the key in the json object. `result = null` still counts as present. In fact the spec explicitly states

>  If a request doesn’t provide a result value the receiver of a request still needs to return a response message to conform to the JSON RPC specification. The result property of the ResponseMessage should be set to null in this case to signal a successful request.

Switching to an `Either` would definitly be the right thing to do if we were starting from scratch (in my opinion). However, at this point it would be a breaking change and I’m not quite sure that’s worth it (but I’m not opposed to it, for our codebase it would be an easy change). However, it’s also somewhat orthogonal to this PR. You still need to handle `result = null` correctly.

alanz:
I know we have had some subtleties in hie around this, in terms of client expectations, and the way it is now is as a result of that.

So we need to be careful about exactly what we change and how.

cocreature:
@alanz By “this” you mean not using some kind of sumtype or the `FromJSON` instance? The current `FromJSON` instance is just broken as far as I can tell so I would hope that HIE does not rely on it.

alanz:
@cocreature I just scanned the issues for hie and haskell-lsp, it seems the problem related to the Hover message, which is now resolved.

So this change is probably ok then. Worst case we get a live issue which gives us something concrete to look at.

cocreature:
Happy to help fix issues caused by this if there are any.

PR #173, opened 2019-06-09
Use attoparsec to parse message headers incrementally

cocreature:
Previously, we were repeatedly reading one character, appending it to
the end and trying to parse until parsec returned a result. This is
needlessly wasteful and while it’s probably not a bottleneck, I think
the incremental parsing with attoparsec is also easier to understand
so it’s a win both in readability and performance.

### DIFF
--- a/src/Language/Haskell/LSP/Core.hs
+++ b/src/Language/Haskell/LSP/Core.hs
@@ -497,8 +497,8 @@ defaultProgressData = ProgressData 0 Map.empty
 -- ---------------------------------------------------------------------
 
 handleMessage :: (Show c) => InitializeCallback c
-              -> TVar (LanguageContextData c) -> BSL.ByteString -> BSL.ByteString -> IO ()
-handleMessage dispatcherProc tvarDat contLenStr jsonStr = do
+              -> TVar (LanguageContextData c) -> BSL.ByteString -> IO ()
+handleMessage dispatcherProc tvarDat jsonStr = do
   {-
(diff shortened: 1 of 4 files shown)


src/Language/Haskell/LSP/Core.hs
```diff
@@ -497,8 +497,8 @@ defaultProgressData = ProgressData 0 Map.empty
 -- ---------------------------------------------------------------------
 
 handleMessage :: (Show c) => InitializeCallback c
-              -> TVar (LanguageContextData c) -> BSL.ByteString -> BSL.ByteString -> IO ()
-handleMessage dispatcherProc tvarDat contLenStr jsonStr = do
+              -> TVar (LanguageContextData c) -> BSL.ByteString -> IO ()
```
bubba: I was looking at this function today, and I don't think it's meant to be exported and it's only used by Control.hs. Not a part of this PR, but should this be moved into its own module/moved into Control.hs?

cocreature: I’m generally a fan of exposing internals at least in some `.Internal` module since I find it quite frustrating when I find the code that I’m looking for but it’s not exposed. That said, for this particular function I don’t see a lot of usecases so I’d be fine with not exposing it. It is however used from `WorkspaceFoldersSpec`.

alanz: I agree with @cocreature , it is very frustrating not to be able to access internal functions.

PR #174, opened 2019-06-09
 Document textDocumentDidChange handler to mention VFS and suppress warnings

bubba:
@EggBaconAndSpam mentioned during his GSOC work on dhall-lsp-server that it wasn't obvious that the VFS exists! We should probably document this better. Later on I will try and update the example to use VFS


alanz:
@bubba we still need to add a note about needing to provide trivial handlers for the change notification messages?

bubba:
@alanz @EggBaconAndSpam I was thinking we could probably just provide a default nop handler?

alanz:
@bubba I agree

bubba:
@alanz done

PR #176, opened 2019-06-12
Normalize URIs to avoid issues with percent encoding

cocreature:
Previously, haskell-lsp treated URIs that only differ in
percent-encoding. This does not comply with the URI spec and it can
also lead to really confusing errors where e.g., a VFS lookup fails
due to using a slightly different percent-encoding.

### DIFF
--- a/example/Main.hs
+++ b/example/Main.hs
@@ -104,7 +104,7 @@ reactorSend msg = do
 
 -- ---------------------------------------------------------------------
 
-publishDiagnostics :: Int -> J.Uri -> J.TextDocumentVersion -> DiagnosticsBySource -> R () ()
+publishDiagnostics :: Int -> J.NormalizedUri -> J.TextDocumentVersion -> DiagnosticsBySource -> R () ()
 publishDiagnostics maxToPublish uri v diags = do
   lf <- ask
   liftIO $ (Core.publishDiagnosticsFunc lf) maxToPublish uri v diags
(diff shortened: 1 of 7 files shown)


alanz:
Can you add a test that shows that two differently encoded urls are treated the same?

cocreature:
Good idea, I’ve added a QuickCheck test that checks that after applying `toNormalizedUri` things are equal.

alanz:
LGTM, we can merge if the rest are happy

example/Main.hs
```diff
@@ -183,15 +183,16 @@ reactor lf inp = do
...
 
       -- -------------------------------
 
       HandlerRequest (NotDidChangeTextDocument notification) -> do
-        let doc :: J.Uri
+        let doc :: J.NormalizedUri
             doc  = notification ^. J.params
                                  . J.textDocument
                                  . J.uri
+                                 . to J.toNormalizedUri
```
bubba: For now, are we just normalising didchangetextdocument notifications?

cocreature: We are normalizing everything in the VFS and everything in the DiagnosticsStore.

bubba: This sounds good to me. How breaking would it be to make this the default behaviour for `URL`?
Could we also define a custom `Eq` instance rather than have a smart constructor to preserve roundtripping?

PR #177, opened 2019-06-27
Changed the initial callbacks type to also capture initial config

lorenzo:
The initial configuration sent by the client was being ignored. This
changes the types and handlers to read initial configuration and
remember it in the LangaugeContext value

### DIFF
--- a/src/Language/Haskell/LSP/Core.hs
+++ b/src/Language/Haskell/LSP/Core.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE GADTs               #-}
 {-# LANGUAGE MultiWayIf          #-}
 {-# LANGUAGE BinaryLiterals      #-}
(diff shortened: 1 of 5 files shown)


src/Language/Haskell/LSP/Core.hs
```diff
@@ -638,10 +698,6 @@ initializeRequestHandler' (_configHandler,dispatcherProc) mHandler tvarCtx req@(
     atomically $ modifyTVar' tvarCtx (\c -> c { resWorkspaceFolders = wfs })
 
     ctx0 <- readTVarIO tvarCtx
-
-    -- capture initialize request
-    captureFromClient (ReqInitialize req) (resCaptureFile ctx0)
```
alanz: Why has this been removed?

lorenzo: It's already captured by the wrapper to this function. Is there a problem in capturing twice? I'm proceeding from the understanding that it is a problem indeed

Issue #134, opened 2018-08-28
Add 3.11 and 3.12 support


Bubba:

EggBaconAndSpam:
I would love support for this one in particular:

> ## 3.13
> 
>     * [ ]   Add support for file and folder operations (create, rename, move) to workspace edits.

bubba:
@EggBaconAndSpam I believe we are yet to add workspace support first. I think we have the types defined, but yes this would be a great feature to have! It might be nice to have a separate issue for this since there’s probably a lot of discussion needed around how the VFS/api will work around it, assuming that we’re adding deeper integration than just handlers.

EggBaconAndSpam:
My use-case is the following: I would like to be able to create files via the "workspace/applyEdit" command; as far as I understand it *workspaces* as a concept are largely orthogonal to this feature. Currently the corresponding haskell-lsp-types definition only covers edits to existing files:


(From [WorkspaceEdit.hs](https://github.com/alanz/haskell-lsp/blob/d00f83700cdc681f65015c7343206c134919d15f/haskell-lsp-types/src/Language/Haskell/LSP/Types/WorkspaceEdit.hs#L146-L151))

```
data WorkspaceEdit =
  WorkspaceEdit
    { _changes         :: Maybe WorkspaceEditMap
    , _documentChanges :: Maybe (List TextDocumentEdit)
    } deriving (Show, Read, Eq)
```
Which corresponds to `documentChanges?: TextDocumentEdit[];`, while the new spec extends this to `documentChanges?: (TextDocumentEdit[] | (TextDocumentEdit | CreateFile | RenameFile | DeleteFile)[]);`. I propose something along these lines:

```
data DocumentChange = DCEdit TextDocumentEdit
                    | DCCreate CreateFile
                    | DCRename RenameFile
                    | DCDelete DeleteFile


data WorkspaceEdit =
  WorkspaceEdit
    { _changes         :: Maybe WorkspaceEditMap
    , _documentChanges :: Maybe (List DocumentChange)
    } deriving (Show, Read, Eq)
```

thomasjm:
I just made a pair of PRs for 3.11 and 3.12 support. They're not really adequately tested yet but hopefully this is a step in the right direction.

jneira:
Updating @theobat tracking:

##  3.3 
- [ ]  Added support for CompletionContext
- [ ]  Added support for MarkupContent (done for `Hover`, still needed for some others)

## 3.13
[...]

PR #178, opened 2019-07-15
Add support CodeActionOptions (added in LSP spec 3.11.0)

thomasjm:
Cf. issue #134


src/Language/Haskell/LSP/Core.hs
```diff
@@ -846,7 +847,7 @@ initializeRequestHandler' onStartup mHandler tvarCtx req@(J.RequestMessage _ ori
               , J._documentHighlightProvider        = supported (documentHighlightHandler h)
               , J._documentSymbolProvider           = supported (documentSymbolHandler h)
               , J._workspaceSymbolProvider          = supported (workspaceSymbolHandler h)
-              , J._codeActionProvider               = supported (codeActionHandler h)
+              , J._codeActionProvider               = codeActionProvider o
```
mpickering: This line of the patch is incorrect. It should be implemented like `typeDefinitionProvider`.

PR #179, opened 2019-07-15
Add support for textDocument/prepareRename (added in LSP 3.12)

thomasjm:
Cf. issue #134

### DIFF
--- a/haskell-lsp-types/src/Language/Haskell/LSP/Types/DataTypesJSON.hs
+++ b/haskell-lsp-types/src/Language/Haskell/LSP/Types/DataTypesJSON.hs
@@ -2588,6 +2588,37 @@ type RenameResponse = ResponseMessage WorkspaceEdit
 
 -- ---------------------------------------------------------------------
 {-
+Prepare Rename Request
+
+Since version 3.12.0
+
+The prepare rename request is sent from the client to the server to setup
+and test the validity of a rename operation at a given location.
+
+Request:
+
+    method: ‘textDocument/prepareRename’
+    params: TextDocumentPositionParams
+
+Response:
+
+    result: Range | { range: Range, placeholder: string } | null describing
+            the range of the string to rename and optionally a placeholder
+            text of the string content to be renamed. If null is returned
+            then it is deemed that a ‘textDocument/rename’ request is not
+            valid at the given position.
+    error: code and message set in case an exception happens during the
+           prepare rename request.
+
+-}
+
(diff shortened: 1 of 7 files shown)


haskell-lsp-types/src/Language/Haskell/LSP/Types/DataTypesJSON.hs
```diff
@@ -2586,6 +2611,37 @@ deriveJSON lspOptions ''RenameParams
...
+            valid at the given position.
+    error: code and message set in case an exception happens during the
+           prepare rename request.
+
+-}
+
+-- {\"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"textDocument/rename\",\"params\":{\"textDocument\":{\"uri\":\"file:///home/alanz/mysrc/github/alanz/haskell-lsp/src/HieVscode.hs\"},\"position\":{\"line\":37,\"character\":17},\"newName\":\"getArgs'\"}}
+
+type PrepareRenameRequest  = RequestMessage ClientMethod TextDocumentPositionParams Range
+type PrepareRenameResponse = ResponseMessage Range
```
bubba: Should we handle the case where it can return a range and a placeholder?

bubba:
Can you update the helper functions inside Capabilities.hs to reflect that this is available in 3.12?

haskell-lsp-types/src/Language/Haskell/LSP/Types/DataTypesJSON.hs
thomasjm: Yes, I'll add a type for that

thomasjm:
Ah I see -- please double check it looks good now.

haskell-lsp-types/src/Language/Haskell/LSP/Types/DataTypesJSON.hs
```diff
@@ -307,6 +307,31 @@ data DocumentLinkOptions =
...
+         */
+        prepareProvider?: boolean;
+}
+-}
+
+data RenameOptions =
+  RenameOptionsStatic Bool
+  | RenameOptions
+    { -- |Renames should be checked and tested before being executed.
+      _prepareProvider :: Maybe Bool
```
lorenzo: Would you mind using this other comment notation?

```haskell
data RenameOptions =
	  RenameOptionsStatic Bool
	  | RenameOptions
	    {  _prepareProvider :: Maybe Bool -- ^ Renames should be checked and tested before being executed.
```

haskell-lsp-types/src/Language/Haskell/LSP/Types/DataTypesJSON.hs
thomasjm: Just added the type salad for this

PR #182, opened 2019-08-25
Fix compilation for PrepareRename

alanz:
The merge for #179 failed, and there were some bits missing for the
full implementation of the feature.


alanz:
cc @thomasjm

thomasjm:
Ah, thanks for fixing that. I only use `haskell-lsp-types` so didn't notice `haskell-lsp` needed updating.

PR #188, opened 2019-10-14
Add ProgressToken to all missing data types

cocreature:
This is based on @bubba’s branch. I’ve gone through all types in the specification and added it so this should be exhaustive (ignoring types that are completely missing in `haskell-lsp` atm, e.g., `PrepareRenameParams`).

fixes #186

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,5 +1,9 @@
 # Revision history for haskell-lsp
 
+## 0.17.0.0 -- unreleased
+
+* Update progress reporting to match the LSP 3.15 specification.
+
 ## 0.16.0.0 -- 2019-09-07
 
 * Add support for CodeActionOptions (@thomasjm)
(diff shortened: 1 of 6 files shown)


alanz:
Is it time to remove GHC 8.0.2 as an officially supported version? And to add GHC 8.6.5

cocreature:
@alanz I’m fine with removing `8.0` support but note that the CI failure is due to a stack bug that has already been fixed but not been released, not because `haskell-lsp` is broken on `8.0`, see https://github.com/commercialhaskell/stack/pull/4955

alanz:
And the build passed on master? https://circleci.com/workflow-run/dd4f4c1f-9af6-4c3d-80a1-57e68ea8c67c

cocreature:
I guess stack might only try to call `ghc-pkg unregister` if the cache is in a certain state and the cache on master might be different? Not quite sure. Maybe a sign that we should switch CI to `cabal new-build` :slightly_smiling_face:

bubba:
@alanz @cocreature I’d just probably just nuke the cache

PR #189, opened 2019-10-14
Ensure that ResponseMessage has either a result or an error

cocreature:
Without this, it can happen that a request is parsed as a response
with both _result and _error set to Nothing. This has caused issues
for me with lsp-test which listens for a response message with the
same id after sending a request but actually ended up interpreting a
request send from the server as a response.

### DIFF
--- a/haskell-lsp-types/src/Language/Haskell/LSP/Types/Message.hs
+++ b/haskell-lsp-types/src/Language/Haskell/LSP/Types/Message.hs
@@ -3,10 +3,12 @@
 {-# LANGUAGE TemplateHaskell            #-}
 module Language.Haskell.LSP.Types.Message where
 
+import           Control.Monad
 import qualified Data.Aeson                                 as A
 import           Data.Aeson.TH
 import           Data.Aeson.Types
 import           Data.Hashable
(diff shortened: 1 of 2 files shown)


cocreature:
This seems to be another hint at the fact that we should change this to something like `Either a ResponseError` but that would be a breaking change so I would at least do that in a separate PR whereas this is only breaking if you rely on the behavior on invalid `json-rpc` messages.

alanz:
LGTM

bubba:
> This has caused issues
for me with lsp-test which listens for a response message with the
same id after sending a request but actually ended up interpreting a
request send from the server as a response.

Out of curiosity, how come there was a request being sent from the server with the same ID as another request?

cocreature:
@bubba Request ids coming from requests from the server are separate from request ids coming from the client. At least that’s the case in `hasell-lsp`. I tried to figure out if that conforms to the spec and didn’t get a clear answer but the underlying json-rpc protocol seems to assume that things are undirectional so if both sides can send requests you effectively have two streams and then it makes sense that you can get the same request ids for requests coming from different sides.

PR #192, opened 2019-10-16
Use Code Action provider based on option and provider

fendor:
Based on the discussions with @alanz and @mpickering

### DIFF
--- a/src/Language/Haskell/LSP/Core.hs
+++ b/src/Language/Haskell/LSP/Core.hs
@@ -844,6 +844,10 @@ initializeRequestHandler' onStartup mHandler tvarCtx req@(J.RequestMessage _ ori
           static (Just d) _ = Just d
           static _ (Just _) = Just (J.GotoOptionsStatic True)
           static _ Nothing  = Nothing
+          
+          static' (Just d) (Just _) = Just d
+          static' _ (Just _) = Just (J.CodeActionOptionsStatic True)
+          static' _ _  = Nothing
 
           sync = case textDocumentSync o of
                   Just x -> Just (J.TDSOptions x)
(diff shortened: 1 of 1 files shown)


alanz:
LGTM

src/Language/Haskell/LSP/Core.hs
```diff
@@ -844,6 +844,9 @@ initializeRequestHandler' onStartup mHandler tvarCtx req@(J.RequestMessage _ ori
           static (Just d) _ = Just d
           static _ (Just _) = Just (J.GotoOptionsStatic True)
           static _ Nothing  = Nothing
+          static' (Just d) (Just _) = Just d
```
jneira: This one is already covered by `static (Just d) _ = Just d`, no?

src/Language/Haskell/LSP/Core.hs
```diff
@@ -844,6 +844,10 @@ initializeRequestHandler' onStartup mHandler tvarCtx req@(J.RequestMessage _ ori
           static (Just d) _ = Just d
           static _ (Just _) = Just (J.GotoOptionsStatic True)
           static _ Nothing  = Nothing
+          
+          static' (Just d) (Just _) = Just d
+          static' _ (Just _) = Just (J.CodeActionOptionsStatic True)
+          static' _ _  = Nothing
```
mpickering: You could modify `static` to take an extra default argument, then you can use it for both the cases.

alanz: I am happy with either option.

fendor: Cases are not identical, if I see that correctly.
`static'` only returns `Just d` if both provider and handler are set, while `static` only requires the provider?

mpickering: Yeah, but why are they different?

jneira:
oh, sorry, i misreaded the functions names and i thought my comment had no sense anymore, but yeah, separating them (or given them different meaningful names) could help to avoid the mistake

PR #196, opened 2019-10-27
Rework Core.Options and infer all server capabilities from handlers

### LINKED ISSUE #195: Make sure all capabilites are determined from handlers
In particular, rename and implementation providers.

See https://github.com/haskell/haskell-ide-engine/pull/1427

bubba:
Previously we were using a mixture of inferring server capabilities based on wether or not there was a handler registered for it, and wether or not the user of the library had set options for it. 

This changes it to the former, so that Core.Options only contains options that are needed to tweak client behaviour, and all the actual capabilities are determined based on wether or not there is a handler set.

It also adds some stricter types, and throws errors if the user of the library forgets to set some mandatory options when a specific handler is set, for example executeCommand and documentOnTypeFormatting.

Closes #195

### DIFF
--- a/src/Language/Haskell/LSP/Core.hs
+++ b/src/Language/Haskell/LSP/Core.hs
@@ -33,7 +33,7 @@ import           Control.Concurrent.Async
 import qualified Control.Exception as E
 import           Control.Monad
 import           Control.Monad.IO.Class
-import           Control.Lens ( (<&>), (^.) )
+import           Control.Lens ( (<&>), (^.), (^?), _Just )
 import qualified Data.Aeson as J
 import qualified Data.ByteString.Lazy as BSL
 import qualified Data.ByteString.Lazy.Char8 as B
@@ -840,15 +840,17 @@ initializeRequestHandler' onStartup mHandler tvarCtx req@(J.RequestMessage _ ori
         sendResponse tvarCtx $ RspError $ makeResponseError (J.responseId origId) errResp
 
       Nothing -> do
-        let capa = serverCapabilities (resOptions ctx) (resHandlers ctx)
+        let capa = serverCapabilities (getCapabilities params) (resOptions ctx) (resHandlers ctx)
             -- TODO: wrap this up into a fn to create a response message
             res  = J.ResponseMessage "2.0" (J.responseId origId) (Just $ J.InitializeResponseCapabilities capa) Nothing
 
         sendResponse tvarCtx $ RspInitialize res
 
 -- | Infers the capabilities based on registered handlers, and sets the appropriate options.
-serverCapabilities :: Options -> Handlers -> J.InitializeResponseCapabilitiesInner
-serverCapabilities o h =
+-- A provider should be set to Nothing if the server does not support it, unless it is a
+-- static option.
+serverCapabilities :: C.ClientCapabilities -> Options -> Handlers -> J.InitializeResponseCapabilitiesInner
+serverCapabilities clientCaps o h =
   J.InitializeResponseCapabilitiesInner
     { J._textDocumentSync                 = sync
     , J._hoverProvider                    = supported (hoverHandler h)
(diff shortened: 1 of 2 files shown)


src/Language/Haskell/LSP/Core.hs
```diff
@@ -831,67 +840,94 @@ initializeRequestHandler' onStartup mHandler tvarCtx req@(J.RequestMessage _ ori
...
+-- | Infers the capabilities based on registered handlers, and sets the appropriate options.
+serverCapabilities :: Options -> Handlers -> J.InitializeResponseCapabilitiesInner
+serverCapabilities o h =
+  J.InitializeResponseCapabilitiesInner
+    { J._textDocumentSync                 = sync
+    , J._hoverProvider                    = supported (hoverHandler h)
+    , J._completionProvider               = completionProvider
+    , J._signatureHelpProvider            = signatureHelpProvider
+    , J._definitionProvider               = supported (definitionHandler h)
+    , J._typeDefinitionProvider           = Just $ J.GotoOptionsStatic $ isJust $ typeDefinitionHandler h
```
alanz: I am not sure this duplicates the current behaviour, which can generate `Nothing` if the handler is not set.

bubba: I think it's better to be more explicit about this and return `False` in this case. If we don't have a handler we know we won't be able to support it, and it's not exactly clearly defined in the specification what `null` means:


    /**
	 * The server provides Goto Type Definition support.
	 *
	 * @since 3.6.0
	 */
	typeDefinitionProvider?: boolean | TypeDefinitionOptions | TypeDefinitionRegistrationOptions;

alanz: Makes sense

src/Language/Haskell/LSP/Core.hs
```diff
@@ -95,29 +96,37 @@ data ProgressData = ProgressData { progressNextId :: !Int
...
+    -- | The characters that trigger signature help automatically.
+    , signatureHelpTriggerCharacters   :: Maybe [Char]
+    -- | List of characters that re-trigger signature help.
+    -- These trigger characters are only active when signature help is already showing. All trigger characters
+    -- are also counted as re-trigger characters.
+    , signatureHelpRetriggerCharacters :: Maybe [Char]
+    -- | CodeActionKinds that this server may return.
+    -- The list of kinds may be generic, such as `CodeActionKind.Refactor`, or the server
+    -- may list out every specific kind they provide.
+    , codeActionKinds                  :: Maybe [J.CodeActionKind]
```
cocreature: I think the API for this isn’t quite right (but that’s not new to this PR):
The LSP spec states

> The `CodeActionOptions` return type is only valid if the client signals code action literal support via the property  `textDocument.codeAction.codeActionLiteralSupport`.

However, the current API requires you to specify the options before you can inspect the capabilities so you can’t actually specify this properly.

src/Language/Haskell/LSP/Core.hs
```diff
@@ -831,67 +840,94 @@ initializeRequestHandler' onStartup mHandler tvarCtx req@(J.RequestMessage _ ori
...
+
+    completionProvider
+      | isJust $ completionHandler h = Just $
+          J.CompletionOptions
+            (Just $ isJust $ completionResolveHandler h)
+            (map singleton <$> completionTriggerCharacters o)
+            (map singleton <$> completionAllCommitCharacters o)
+      | otherwise = Nothing
+
+    codeActionProvider
```
cocreature: This breaks `ghcide` in a way that I don’t know how to fix. We currently set this to `CodeActionOptionsStatic True` and as mentioned above, I can’t set the individual flags properly since I learn the capabilities after I need to specify the options. Maybe for now, we could use something like
```
isJust $ codeActionHandler h = maybe (J.CodeActionOptionsStatic True) (CodeActionOptions . Just) (codeActionKinds o)
```

src/Language/Haskell/LSP/Core.hs
bubba: Good catch. I think we can just return `CodeActionOptionsStatic True` in the case when the client capabilities do not declare support for code action literals, and pass on the code action options in the case that they do

src/Language/Haskell/LSP/Core.hs
bubba: That sounds reasonable. I left it this way as it is possible to return a `CodeActionOptions` with a null `CodeActionKinds` inside it (which doesn't make much sense given the previous comment), so I'm happy enough to default to the static capability if the kinds in the LSP options are set to Nothing

PR #197, opened 2019-11-03
Make explicit GHC 8.6.5 stack file

alanz:


alanz:
It also removes the existing `stack.yaml` (which is identical to `stack-8.8.1.yaml`) so you can have a "poor man's hie-bios" by symlinking a `stack.yaml` to whatever you want to work with.

PR #199, opened 2019-11-06
Virtual file fixes

mpickering:


src/Language/Haskell/LSP/VFS.hs
```diff
@@ -62,35 +67,49 @@ data VirtualFile =
...
     , _text    :: Rope
-    , _tmp_file :: Maybe FilePath
     } deriving (Show)
 
-type VFS = Map.Map J.NormalizedUri VirtualFile
+type VFSMap = Map.Map J.NormalizedUri VirtualFile
+
+data VFS = VFS { vfsMap :: Map.Map J.NormalizedUri VirtualFile
+               , vfsTempDir :: FilePath -- ^ This is where all the temporary files will be written to
+               } deriving Show
```
alanz: I think we should move the `revMap` in here too.

alanz:
Apart from moving the `revMap`. LGTM. We could possibly do that as a separate PR.

src/Language/Haskell/LSP/Core.hs
```diff
@@ -893,7 +916,6 @@ serverCapabilities clientCaps o h =
           J.CompletionOptions
             (Just $ isJust $ completionResolveHandler h)
             (map singleton <$> completionTriggerCharacters o)
-            (map singleton <$> completionAllCommitCharacters o)
```
mpickering: Did I do something wrong to have to change these lines?

src/Language/Haskell/LSP/VFS.hs
mpickering: Done

mpickering:
@alanz I incorporated your changes into this PR (without the MVar)

mpickering:
I somehow messed up the history quite badly by doing a merge. Let me know if you want the history to be cleaned up somehow.

alanz:
LGTM

PR #201, opened 2019-11-14
Preparing 0.18 release

alanz:


alanz:
@bubba when this is accepted, can you deal with the matching release process for `lsp-test`?

PR #204, opened 2019-11-26
Fixes to persistVirtualFile

mpickering:
1. Make it return a `Maybe` when the file does not exist in the VFS
2. Use an monotonically incrementing counter to keep track of the version rather than relying on the LSP version which has no guarantees about monotonicity.

### DIFF
--- a/src/Language/Haskell/LSP/VFS.hs
+++ b/src/Language/Haskell/LSP/VFS.hs
@@ -154,23 +154,24 @@ virtualFileName :: FilePath -> J.NormalizedUri -> VirtualFile -> FilePath
 virtualFileName prefix uri (VirtualFile ver _) =
   prefix </> show (hash (J.fromNormalizedUri uri)) ++ "-" ++ show ver ++ ".hs"
 
-persistFileVFS :: VFS -> J.NormalizedUri -> (FilePath, IO ())
+-- | Write a virtual file to a temporary file if it exists in the VFS.
+persistFileVFS :: VFS -> J.NormalizedUri -> Maybe (FilePath, IO ())
 persistFileVFS vfs uri =
   case Map.lookup uri (vfsMap vfs) of
-    Nothing -> error ("File not found in VFS: " ++ show uri ++ show vfs)
+    Nothing -> Nothing
     Just vf@(VirtualFile _v txt) ->
       let tfn = virtualFileName (vfsTempDir vfs) uri vf
           action = do
             exists <- doesFileExist tfn
             unless exists (writeFile tfn (Rope.toString txt))
-      in (tfn, action)
+      in Just (tfn, action)
 
 -- ---------------------------------------------------------------------
 
 closeVFS :: VFS -> J.DidCloseTextDocumentNotification -> (VFS, [String])
 closeVFS vfs (J.NotificationMessage _ _ params) =
   let J.DidCloseTextDocumentParams (J.TextDocumentIdentifier uri) = params
-  in (updateVFS (Map.delete (J.toNormalizedUri uri)) vfs,[])
+  in (updateVFS (Map.delete (J.toNormalizedUri uri)) vfs,["Closed: " ++ show uri])
 
 -- ---------------------------------------------------------------------
 {-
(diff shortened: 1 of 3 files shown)


src/Language/Haskell/LSP/VFS.hs
```diff
@@ -151,26 +165,27 @@ changeFromServerVFS initVfs (J.RequestMessage _ _ _ params) = do
 
 -- ---------------------------------------------------------------------
 virtualFileName :: FilePath -> J.NormalizedUri -> VirtualFile -> FilePath
-virtualFileName prefix uri (VirtualFile ver _) =
-  prefix </> show (hash (J.fromNormalizedUri uri)) ++ "-" ++ show ver ++ ".hs"
+virtualFileName prefix uri (VirtualFile _ file_ver _) =
+  prefix </> show (hash (J.fromNormalizedUri uri)) ++ "-" ++ show file_ver ++ ".hs"
```
alanz: I think it would be good to put the original filename in here too, makes it easier to rummage around in the temp directory and identify the file.  Preferably as a prefix.

bubba: Does `uri` not contain the original file name?

src/Language/Haskell/LSP/VFS.hs
```diff
@@ -148,26 +162,27 @@ changeFromServerVFS initVfs (J.RequestMessage _ _ _ params) = do
...
-    Just vf@(VirtualFile _v txt) ->
+    Nothing -> Nothing
+    Just vf ->
       let tfn = virtualFileName (vfsTempDir vfs) uri vf
           action = do
             exists <- doesFileExist tfn
-            unless exists (writeFile tfn (Rope.toString txt))
-      in (tfn, action)
+            unless exists (writeFile tfn (Rope.toString (_text vf)))
+      in Just (tfn, action)
```
bubba: Do these files ever get cleaned up? I feel as if this function should be in CPS style with a handler that passes the filepath and deletes it afterwards

src/Language/Haskell/LSP/VFS.hs
bubba: Does the uri not contain the original file name?

src/Language/Haskell/LSP/VFS.hs
```diff
@@ -17,6 +17,9 @@ module Language.Haskell.LSP.VFS
   (
     VFS(..)
```
bubba: Should we hide the fields of the VFS record?

mpickering: My opinion is no, it would be very annoying for someone if they had to add a missing function themselves.

src/Language/Haskell/LSP/VFS.hs
mpickering: They all get cleaned up at he end when the main process exits. I suppose they could be cleaned up individually using the refactoring you suggest but there doesn't seem to be much benefit.

src/Language/Haskell/LSP/VFS.hs
mpickering: Added.

src/Language/Haskell/LSP/VFS.hs
```diff
@@ -163,7 +163,9 @@ changeFromServerVFS initVfs (J.RequestMessage _ _ _ params) = do
 -- ---------------------------------------------------------------------
 virtualFileName :: FilePath -> J.NormalizedUri -> VirtualFile -> FilePath
 virtualFileName prefix uri (VirtualFile _ file_ver _) =
-  prefix </> show (hash (J.fromNormalizedUri uri)) ++ "-" ++ show file_ver ++ ".hs"
+  let uri_raw = J.fromNormalizedUri uri
+      basename = maybe "" takeFileName (J.uriToFilePath uri_raw)
+  in prefix </> show (hash uri_raw) ++ "-" ++ show file_ver ++ "-" ++ basename ++ ".hs"
```
alanz: Thanks, but a minor nit.  If I am viewing the temp directory sorted by filename, different versions of the same file are going to be scattered all over the place. It may be best to put the hash last, just before the extension.

And for bonus points left pad the version number with zeros, e.g. `001`.

PR #213, opened 2020-01-25
Move "Semigroup WorkspaceEdit" instance

sheaf:
Declaring the instance after the TH splice caused GHC to not be able to see the Semigroup instance needed for the Monoid instance.


alanz:
@bubba I guess it makes sense to have a new release, this is quite an important feature, GHC 8.10

PR #214, opened 2020-01-27
Store a hash in a NormalizedUri

mpickering:
NormalizedUris are supposed to be used as keys in maps. Using a HashMap
with a cached key is far faster than using a Map with the slow ordering
instance.

In particular, ghcide was hammering updateDiagnostics and the `compare`
function for `NormalizedUri` was dominating the profile.

### DIFF
--- a/haskell-lsp-types/src/Language/Haskell/LSP/Types/Uri.hs
+++ b/haskell-lsp-types/src/Language/Haskell/LSP/Types/Uri.hs
@@ -25,7 +25,7 @@ instance NFData Uri
 --
 -- If you care about performance then you should use a hash map. The keys
 -- are cached in order to make hashing very fast.
-data NormalizedUri = NormalizedUri Int !Text
+data NormalizedUri = NormalizedUri !Int !Text
   deriving (Read,Show,Generic, Eq)
 
 -- Slow but compares paths alphabetically as you would expect.
--- a/haskell-lsp-types/src/Language/Haskell/LSP/Types/Uri.hs
+++ b/haskell-lsp-types/src/Language/Haskell/LSP/Types/Uri.hs
@@ -22,17 +22,27 @@ instance NFData Uri
 -- | When URIs are supposed to be used as keys, it is important to normalize
 -- the percent encoding in the URI since URIs that only differ
 -- when it comes to the percent-encoding should be treated as equivalent.
-newtype NormalizedUri = NormalizedUri Text
-  deriving (Eq,Ord,Read,Show,Generic,Hashable)
+--
+-- If you care about performance then you should use a hash map. The keys
+-- are cached in order to make hashing very fast.
+data NormalizedUri = NormalizedUri Int !Text
+  deriving (Read,Show,Generic, Eq)
+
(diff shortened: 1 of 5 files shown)


haskell-lsp-types/src/Language/Haskell/LSP/Types/Uri.hs
```diff
@@ -22,17 +22,27 @@ instance NFData Uri
...
+instance Hashable NormalizedUri where
+  hash (NormalizedUri h _) = h
 
 toNormalizedUri :: Uri -> NormalizedUri
-toNormalizedUri uri =
-    NormalizedUri $ T.pack $ escapeURIString isUnescapedInURI $ unEscapeString $ T.unpack t
+toNormalizedUri uri = NormalizedUri (hash norm) norm
   where (Uri t) = maybe uri filePathToUri (uriToFilePath uri)
         -- To ensure all `Uri`s have the file path like the created ones by `filePathToUri`
+        norm = T.pack $ escapeURIString isUnescapedInURI $ unEscapeString $ T.unpack t
```
alanz: given this is just a value for indexing, maybe we should skip the T.pack, and store a string?

mpickering: Hmm, that would increase the memory footprint quite a bit. I wold prefer to implement the normalisation logic directly on text.

haskell-lsp-types/src/Language/Haskell/LSP/Types/Uri.hs
```diff
@@ -22,17 +22,27 @@ instance NFData Uri
 -- | When URIs are supposed to be used as keys, it is important to normalize
 -- the percent encoding in the URI since URIs that only differ
 -- when it comes to the percent-encoding should be treated as equivalent.
-newtype NormalizedUri = NormalizedUri Text
-  deriving (Eq,Ord,Read,Show,Generic,Hashable)
+--
+-- If you care about performance then you should use a hash map. The keys
+-- are cached in order to make hashing very fast.
+data NormalizedUri = NormalizedUri Int !Text
```
cocreature: Is the hash intentionally non-strict? I guess it saves you a bit if you don’t care about it but if you do care about it, making it strict so it can be unboxed seems benefitiial.

pepeiborra: You want to delay the hashing until it is actually used, which may never happen. So it makes sense to keep it non-strict

pepeiborra:
Text has an efficient `Hashable` instance. Do you need both changes, or maybe just the switch to `HashMap`?

mpickering:
The hashing method still accounts for 2.8% of runtime if you don't cache it so I would prefer to leave it as it is for now.

haskell-lsp-types/src/Language/Haskell/LSP/Types/Uri.hs
mpickering: The hash is intentionally non-strict for the reasons that Pepe suggests.

PR #223, opened 2020-02-20
Add diagnostic and completion tags

SquidDev:
These appear in the [3.15 spec][spec], and provide a way to provide additional metadata about diagnostics and completion items. For instance, the "Unnecessary" tag says this diagnostic warns about unused code or definitions.

### Example
Just demonstrating the "unused" tag being rendered by VS Code.
image

[spec]: https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/


SquidDev:
Thank you!

PR #224, opened 2020-02-21
Add NormalizedFilePath from ghcide

### LINKED ISSUE #212: Refactor file uris to make possible skip normalization
* After #205 normalization could happen up to 2 times: in `filePathToUri` and `toNormalizeUri`, the last one to ensure that file `NormalizedUri`'s from other sources (like Aeson) are always normalized (and fix a bug in hie)
* Moreover, if client code already has ensured that filepaths are normalized it suffer the penalty of two additional normalizations:
  * It happened in ghcid: https://github.com/digital-asset/ghcide/pull/303 with a notable perfomance degradation (it is called many times)
  * So it was neccesary to inline the code of `filePathToUri`to avoid the extra normalization
* To avoid this, `filePathToUri` should remove the normalization, moving it to `normalizeUri` and maybe to another function `filePathToNormalizedUri :: FilePath -> NormalizedUri`

jneira:
* To track in types we have a normalized path and it is sure to conver to normalized uri without an extra normalization step
* Fixes #212 and let ghcide not worry about its specofoc implementation
* The implementation only calls `FP.normalise` once in `fromNormalizedFilePath . uriToNormalizedFilePath . normalizedFilePathToUri . toNormalizedFilePath`, like the actual ghcide impl so i hope the performance will be not degraded
  * it would be useful to have a benchmark over it though
* Significant changes:
  * *platform\* functions are not public any more* cause they are a implementation detail that should not be exposed
  * *the hash of NormalizedFilePath is equal now to the hash of NormalizedUri*: imo they *must* match for all cases and we avoid an additional hash call (/cc @cocreature)
  * *tests dont check both os's anymore* so the actual ci will not tests windows uris. I want to add windows ci soon so the tests will be recoved
    * The test suit passes locally in my windows systems
  * `uriTofilePath` and `filePathToUri` are now deprecated, with the goal of replace them with the new `NormalizedFilePath` functions. They should be removed in the next breaking release after this one.
* Caveats: [i've kept the ghcide corner case for empty file paths](https://github.com/alanz/haskell-lsp/pull/224/files#diff-c671434724ea00cc2090da17fd103d7dR167) but i think it would be nice to not include it in the library. Not sure if it is possible without significant changes in the lib or ghcide code

### DIFF
--- a/haskell-lsp-types/src/Language/Haskell/LSP/Types/Uri.hs
+++ b/haskell-lsp-types/src/Language/Haskell/LSP/Types/Uri.hs
@@ -140,28 +140,28 @@ platformAdjustToUriPath systemOS srcPath
 --
 -- This is one of the most performance critical parts of ghcide, do not
 -- modify it without profiling.
-data NormalizedFilePath = NormalizedFilePath NormalizedUri !Int !FilePath
+data NormalizedFilePath = NormalizedFilePath NormalizedUri !FilePath
     deriving (Generic, Eq, Ord)
 
 instance NFData NormalizedFilePath
 
 instance Binary NormalizedFilePath where
-  put (NormalizedFilePath _ _ fp) = put fp
+  put (NormalizedFilePath _ fp) = put fp
   get = do
     v <- Data.Binary.get :: Get FilePath
     return (toNormalizedFilePath v)
 
 instance Show NormalizedFilePath where
-  show (NormalizedFilePath _ _ fp) = "NormalizedFilePath " ++ show fp
+  show (NormalizedFilePath _ fp) = "NormalizedFilePath " ++ show fp
 
 instance Hashable NormalizedFilePath where
-  hash (NormalizedFilePath _ h _) = h
+  hash (NormalizedFilePath uri _) = hash uri
 
 instance IsString NormalizedFilePath where
     fromString = toNormalizedFilePath
 
 toNormalizedFilePath :: FilePath -> NormalizedFilePath
-toNormalizedFilePath fp = NormalizedFilePath nuri (hash nfp) nfp
+toNormalizedFilePath fp = NormalizedFilePath nuri nfp
   where nfp | fp == "" = "" 
             -- ghcide want to keep empty paths instead of normalising them to "."
             | otherwise = FP.normalise fp
@@ -170,13 +170,13 @@ toNormalizedFilePath fp = NormalizedFilePath nuri (hash nfp) nfp
         nuri = NormalizedUri (hash nuriStr) nuriStr
 
 fromNormalizedFilePath :: NormalizedFilePath -> FilePath
-fromNormalizedFilePath (NormalizedFilePath _ _ fp) = fp
+fromNormalizedFilePath (NormalizedFilePath _ fp) = fp
 
 normalizedFilePathToUri :: NormalizedFilePath -> NormalizedUri
-normalizedFilePathToUri (NormalizedFilePath uri _ _) = uri
+normalizedFilePathToUri (NormalizedFilePath uri _) = uri
 
 uriToNormalizedFilePath :: NormalizedUri -> Maybe NormalizedFilePath
 uriToNormalizedFilePath nuri = fmap toNormFP mbFilePath
   where mbFilePath = platformAwareUriToFilePath System.Info.os (fromNormalizedUri nuri) 
         -- This file path is already normalized by construction
-        toNormFP nfp = NormalizedFilePath nuri (hash nfp) nfp
+        toNormFP nfp = NormalizedFilePath nuri nfp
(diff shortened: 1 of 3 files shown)


jneira:
@cocreature ok, i would prefer to keep the `platform*` functions private cause imo they should not be used out of the module but i'll add them to exports with a warning labeled them as "internal"

jneira:
@cocreature ok, i've recovered the tests over the platform aware functions as they was in 2dceca9, cause those functions had been reverted to that version (they are not doing a full normalization now).
Anyway i've added a warning to platform aware functions to make clear they should not be used outside the package.
Otoh i've removed the deprecations over `uriToFilePath` and `filePathToUri` cause `Uri` is being used and it is the one that has the Aeson instances, so replace it is a long term task.

haskell-lsp-types/src/Language/Haskell/LSP/Types/Uri.hs
```diff
@@ -113,3 +142,48 @@ platformAdjustToUriPath systemOS srcPath
...
+instance Hashable NormalizedFilePath where
+  hash (NormalizedFilePath uri _) = hash uri
+
+instance IsString NormalizedFilePath where
+    fromString = toNormalizedFilePath
+
+toNormalizedFilePath :: FilePath -> NormalizedFilePath
+toNormalizedFilePath fp = NormalizedFilePath nuri nfp
+  where nfp | fp == "" = "" 
+            -- ghcide want to keep empty paths instead of normalising them to "."
```
jneira: @cocreature what do you think about remove this corner case? i would prefer to keep it in ghcide instead the lib

haskell-lsp-types/src/Language/Haskell/LSP/Types/Uri.hs
```diff
@@ -113,3 +142,48 @@ platformAdjustToUriPath systemOS srcPath
...
+            | otherwise = FP.normalise fp
+        uriStr = fileScheme <> "//" <> platformAdjustToUriPath System.Info.os nfp
+        nuriStr = T.pack $ escapeURIString isUnescapedInURI $ unEscapeString $ uriStr
```
jneira: @cocreature i think this part is making roundtrip tests fail and i am not sure it is intentional. Consider the repl session:
```
> toNormalizedFilePath "C:\\a#s"
NormalizedFilePath "C:\\a#s"
> normalizedFilePathToUri $ toNormalizedFilePath "C:\\a#s"
NormalizedUri 1827512744 "file:///C:/a#s"
> uriToNormalizedFilePath $ normalizedFilePathToUri $ toNormalizedFilePath "C:\\a#s"
Just NormalizedFilePath "C:\\a"
```
shouldnt `NormalizedUri _"file:///C:/a#s"` be `NormalizedUri _"file:///C:/a%23s"`?
Like the old functions already does:
```
> filePathToUri "C:\a#s"
Uri {getUri = "file:///C:/%07%23s"}
> uriToFilePath $ filePathToUri "C:\\a#s"
Just "C:\\a#s"
```

jneira: I've removed temporary the extra unescape/escape uri step to check if tests pass

jneira:
I've added some changes to make the roundtrip tests pass, to do so i've have to change the function that normalizes the percent encoding

I want to highlight the new test cases:

```haskell
it "converts a file path with reserved uri chars to a normalized URI and back" $ do
    let start = if isWindows then "C:\\" else "/"
    let fp = start ++ "path;part#fragmen?param=val"
    let nuri = toNormalizedUri (filePathToUri fp)
    uriToFilePath (fromNormalizedUri nuri) `shouldBe` Just fp

  it "converts a file path with substrings that looks like uri escaped chars and back" $ do
    let start = if isWindows then "C:\\" else "/"
    let fp = start ++ "ca%C3%B1a"
    let nuri = toNormalizedUri (filePathToUri fp)
    uriToFilePath (fromNormalizedUri nuri) `shouldBe` Just fp
```

Before changing the normalization of percent encoding those tests did not hold and files with `#`, `?`, etc was not properly handled.

To achieve that i had to add another computation to `toNormalizedFilePath`: `stripPrefix "file://"`. 
However i've removed the `T.pack $ escapeURIString isUnescapedInURI $ unEscapeString $ T.unpack t` from `toNormalizedFilePath` so afaiu the performance of ghcide (that uses the last one) should be even better

jneira:
@cocreature sorry for make more changes after your review, but i have a wip pr to adapt ghcide to this changes and i think it worths move the emtpy path case to ghcide cause other downstream packages could have troubles with that corner case.
The pr is here: https://github.com/digital-asset/ghcide/compare/master...jneira:norm-filepath?expand=1

alanz:
@jneira when you and @cocreature are happy with this please merge it.

I think we should make a new haskell-lsp release soon, there have been a number of changes recently.

jneira:
In the light of https://github.com/digital-asset/ghcide/pull/479 i think this pr could be merged as is. I'll do tomorrow if nobody disagree

Issue #206, opened 2019-12-04
Ignore $ messages from VSCode


jinwoo:
We got an internal report about the error below appearing in the output console of VSCode.

```
[Error - 8:24:17 PM] haskell-lsp:no handler for. Object (fromList [("jsonrpc",String "2.0"),("params",Object (fromList [("value",String "off")])),("method",String "$/setTraceNotification")])
```

https://github.com/eclipse/lsp4j/issues/22 and https://github.com/microsoft/language-server-protocol/issues/109 seem to suggest that messages starting with '$' can be ignored by servers. haskell-lsp should probably do that.

bkomuves:
Hi, I believe the above commit does not fix the issue. The problem seems to be that instead of just a method string, the client can send an object with a method field, and this case is not handled. In fact the example in the original report is not handled.

The following change seems to fix the problem for me: Replace lines 436-440 with:


```
    isOptionalNotification req' =
      case (req', json) of
        (NotCustomClient _, jvalue) -> case jvalue of
          J.String method -> "$/" `T.isPrefixOf` method 
          J.Object object -> case HM.lookup "method" object of
            Just jvalue1    -> case jvalue1 of
              J.String method -> "$/" `T.isPrefixOf` method 
              _               -> False
            Nothing         -> False
        _ -> False
```

jinwoo:
Sorry my fix was incomplete, and thanks for the investigation. I’ll create another PR soon. Or if you already have one, please do it.

PR #226, opened 2020-02-24
Correctly fix the problem with '$/' notifications

jinwoo:
The notifications come in the form of a JSON object rather than a JSON
string.

Fixes https://github.com/alanz/haskell-lsp/issues/206.


jinwoo:
I verified in VSCode that spurious errors are displayed with the old code and they are not anymore with this change. The problem can be seen whenever some settings value change in the Settings UI.

jinwoo:
Can this PR be reviewed/merged please?

PR #230, opened 2020-03-03
Add DidChangeWatchedFilesRegistrationOptions

### LINKED ISSUE #229: Add DidChangeWatchedFilesRegistrationOptions
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_didChangeWatchedFiles

It is either not implemented or I cant find it

fendor:
Makes it possible to dynamically register capabilities for watching file changes.

Closes #229

### DIFF
--- a/haskell-lsp-types/src/Language/Haskell/LSP/Types/DataTypesJSON.hs
+++ b/haskell-lsp-types/src/Language/Haskell/LSP/Types/DataTypesJSON.hs
@@ -1094,6 +1094,100 @@ type UnregisterCapabilityRequest = RequestMessage ServerMethod UnregistrationPar
 type UnregisterCapabilityResponse = ResponseMessage ()
 
 -- ---------------------------------------------------------------------
+
+-- /**
+--  * Describe options to be used when registering for file system change events.
+--  */
+-- export interface DidChangeWatchedFilesRegistrationOptions {
+-- 	/**
+-- 	 * The watchers to register.
+-- 	 */
+-- 	watchers: FileSystemWatcher[];
+-- }
+--
+-- export interface FileSystemWatcher {
+-- 	/**
+-- 	 * The  glob pattern to watch.
+-- 	 *
+-- 	 * Glob patterns can have the following syntax:
+-- 	 * - `*` to match one or more characters in a path segment
+-- 	 * - `?` to match on one character in a path segment
+-- 	 * - `**` to match any number of path segments, including none
+-- 	 * - `{}` to group conditions (e.g. `**​/*.{ts,js}` matches all TypeScript and JavaScript files)
+-- 	 * - `[]` to declare a range of characters to match in a path segment (e.g., `example.[0-9]` to match on `example.0`, `example.1`, …)
+-- 	 * - `[!...]` to negate a range of characters to match in a path segment (e.g., `example.[!0-9]` to match on `example.a`, `example.b`, but not `example.0`)
+-- 	 */
+-- 	globPattern: string;
+--
+-- 	/**
+-- 	 * The kind of events of interest. If omitted it defaults
+-- 	 * to WatchKind.Create | WatchKind.Change | WatchKind.Delete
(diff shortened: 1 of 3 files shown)


haskell-lsp-types/src/Language/Haskell/LSP/Types/DataTypesJSON.hs
```diff
@@ -1093,6 +1093,100 @@ type UnregisterCapabilityRequest = RequestMessage ServerMethod UnregistrationPar
...
+      deleteNum = if _watchDelete wk then 4 else 0
+
+instance A.FromJSON WatchKind where
+  parseJSON (A.Number n)
+    | 1 <= n && n <= 7 =
+      let
+        (wkDelete, n1) = if 4 <= n then (True, n-4) else (False, n)
+        (wkChange, n2) = if 2 <= n1 then (True, n1-2) else (False, n1)
+        (wkCreate, _) = if 1 <= n2 then (True, n2-1) else (False, n2)
+      in
```
fendor: Really ugly, but I could not think of a way without introducing loads of types and boilerplate.

bubba: This looks like a bit mask, so a common way would be with a [bitwise and](https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Bits.html#v:.-38-.)

Something like

```haskell
parseJSON (A.Number n) = pure $ WatchKind (n .&. 1) (n .&. 2) (n .&. 4)
```

fendor: Agreed, however, `Scientific` doesnt implement `Bits` :/ we could translate it to Int, but I think that is unsafe.

bubba: Argh I forgot about Javascript numbers. Surely anything thats not an integer is invalid anyway right? And should fail parsing

alanz: Given it should be an integer, you can probably use one of the conversion functions to integer, then bitwise it.

fendor: `testBit` is the function to use!

haskell-lsp-types/src/Language/Haskell/LSP/Types/DataTypesJSON.hs
```diff
@@ -1093,6 +1096,96 @@ type UnregisterCapabilityRequest = RequestMessage ServerMethod UnregistrationPar
...
+    where
+      createNum = if _watchCreate wk then 0x1 else 0x0
+      changeNum = if _watchChange wk then 0x2 else 0x0
+      deleteNum = if _watchDelete wk then 0x4 else 0x0
+
+instance A.FromJSON WatchKind where
+  parseJSON (A.Number n)
+    | Right i <- floatingOrInteger n :: Either Double Int
+    , 0 <= i && i <= 7 =
+        pure $ WatchKind (testBit i 0x0) (testBit i 0x1) (testBit i 0x2)
```
bubba: Should this be 0x1, 0x2, 0x4 for the values in watch kind?

bubba: Never mind, I see it’s the nth bit. LGTM

fendor: It took me three attempts to get it right. That's why there is a jsonSpec test now :)

bubba:
Is this good to merge?

fendor:
Actually, did anyone cross check that somehow if I got it right? Where should I add unit-tests for it?

bubba:
This is just adding a new data type, so I think the round trip unit test should suffice

PR #232, opened 2020-03-07
stack.yaml default file

andys8:
I wanted to give the project a try and figure out the bindings with the example language server. The instructions (`stack install :lsp-hello --flag haskell-lsp:demo`) won't work because there is no `stack.yaml` actually named that way. To have the project working by default with stack an existing `stack.yaml` might be helpful.

Otherwise we could extend the instructions to clarify that a stack-<version>.yaml has to be copied or provided as argument.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,0 +1,12 @@
+resolver: nightly-2020-01-31
+
+packages:
+- '.'
+- ./haskell-lsp-types
+
+extra-deps:
+
+flags: {}
+extra-package-dbs: []
+nix:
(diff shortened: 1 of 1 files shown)


alanz:
This was a deliberate choice. Before hie could support cabal projects, you had to use a stack project, but could not choose a non-standard stack file.  So I would symlink a local `stack.yaml` to whichever one I was working on.

hie-bios / hie now supports cabal, so this is no longer needed.

andys8:
So are we closing the PR or should I add the symlink?

PR #234, opened 2020-03-20
Change ResponseMessage result to be `Either ResponseError a`

### LINKED ISSUE #191: Change response type to contain `Either a ResponseError`
> This seems to be another hint at the fact that we should change this to something like `Either a ResponseError` but that would be a breaking change so I would at least do that in a separate PR whereas this is only breaking if you rely on the behavior on invalid `json-rpc` messages.

_Originally posted by @cocreature in https://github.com/alanz/haskell-lsp/pull/189#issuecomment-541631689_

greenhat:
Close #191 

Todo:
- [x] add JSON roundtrip tests for `ResponseMessage` with valid values for `result` field (`string | number | boolean | object | null`);
- [x] check that invalid JSON (`result = null`(without `error` field) and both of them being absent) is handled properly;
- [x] remove `either` dependency (see https://github.com/alanz/haskell-lsp/pull/234#discussion_r395584136);
- [x] remove prefix in `ResponseMessage` fields (see https://github.com/alanz/haskell-lsp/pull/234#discussion_r395561185);

### DIFF
--- a/haskell-lsp-types/haskell-lsp-types.cabal
+++ b/haskell-lsp-types/haskell-lsp-types.cabal
@@ -58,7 +58,6 @@ library
                      , scientific
                      , text
                      , unordered-containers
-                     , either >= 5.0.1.1
   hs-source-dirs:      src
   default-language:    Haskell2010
 
--- a/haskell-lsp-types/haskell-lsp-types.cabal
+++ b/haskell-lsp-types/haskell-lsp-types.cabal
@@ -58,6 +58,7 @@ library
                      , scientific
                      , text
                      , unordered-containers
+                     , either >= 5.0.1.1
   hs-source-dirs:      src
   default-language:    Haskell2010
 
--- a/haskell-lsp-types/src/Language/Haskell/LSP/Types/Message.hs
+++ b/haskell-lsp-types/src/Language/Haskell/LSP/Types/Message.hs
@@ -424,13 +424,13 @@ deriveJSON lspOptions{ fieldLabelModifier = customModifier } ''ResponseError
 
 data ResponseMessage a =
   ResponseMessage
-    { _rmjsonrpc :: Text
-    , _rmid      :: LspIdRsp
-    , _rmresult  :: Either ResponseError a
+    { _jsonrpc :: Text
+    , _id      :: LspIdRsp
+    , _result  :: Either ResponseError a
     } deriving (Read,Show,Eq)
 
 instance ToJSON a => ToJSON (ResponseMessage a) where
-  toJSON (ResponseMessage { _rmjsonrpc = jsonrpc, _rmid = lspid, _rmresult = result })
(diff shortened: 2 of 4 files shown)


haskell-lsp-types/src/Language/Haskell/LSP/Types/Message.hs
```diff
@@ -380,30 +381,58 @@ deriveJSON lspOptions{ fieldLabelModifier = customModifier } ''ResponseError
...
 data ResponseMessage a =
   ResponseMessage
-    { _jsonrpc :: Text
-    , _id      :: LspIdRsp
-    , _result  :: Maybe a
-    , _error   :: Maybe ResponseError
+    { _rmjsonrpc :: Text
+    , _rmid      :: LspIdRsp
+    , _rmresult  :: Either ResponseError a
     } deriving (Read,Show,Eq)
```
alanz: We have been naming the fields to correspond with the spec, with an underscore to allow idiomatic lens generation.  The `OverloadedRecordFields` extension allows us to do this.

So I think it would be better to stick with the current usage, i.e. get rid of the `rm` bit in the field names.

greenhat: Thanks! Got it. I'll remove them. 
I've added `rm` prefixes because compiler errored on `toJson` :
```haskell
instance ToJSON a => ToJSON (ResponseMessage a) where
  toJSON r = object
    [ "jsonrpc" .= _rmjsonrpc r
    , "id" .= _rmid r
    , "result" .= rightToMaybe (_rmresult r)
    , "error" .= leftToMaybe (_rmresult r)
    ]
``` 
with functions (record fields) being ambigous. I'll find another way.

greenhat:
@alanz Thanks for the review! It's work in progress though(draft PR). I'm not happy with roundtrip serialization tests and I need to clean up my code. :)

haskell-lsp-types/haskell-lsp-types.cabal
```diff
@@ -58,6 +58,7 @@ library
                      , scientific
                      , text
                      , unordered-containers
+                     , either >= 5.0.1.1
```
bubba: Nitpicking here, if we’re only using those two helper functions from either would it be worthwhile to just write them ourselves as a `where` clause? Just to avoid adding more dependencies

greenhat: Thanks! Will do.

haskell-lsp-types/src/Language/Haskell/LSP/Types/Message.hs
greenhat: Done.

haskell-lsp-types/haskell-lsp-types.cabal
greenhat: Done.

greenhat:
I finished with all I had planned. Please review and let me know if I missed anything.

PR #235, opened 2020-03-20
Stop getCompletionPrefix from crashing if beforePos is empty

cocreature:


cocreature:
CI is failing for completely unrelated reasons see #236 for the fix. I’ll rebase once that is merged.

PR #247, opened 2020-05-18
Rework capture logic

bgamari:
The previous mechanism was racy as it opened the capture file
on every event. If two requests came in close succession then one thread
could attempt to open the capture file while it was already open in
another thread.


wz1000:
I think we should merge this

bgamari:
In that case I will mark it as ready for review.

bubba:
I'm happy to merge if you are @wz1000

wz1000:
Tests seem to be failing because of a missing `Semigroup` import.

bubba:
We should drop 8.2 support soon, since hls and ghcide don’t support it anymore

bubba:
Those macOS failures look like something flaky with Azure

PR #249, opened 2020-07-24
Avoid unnecessary normalisation in Binary NormalizedFilePath

cocreature:
imho the Binary instance should be able to assume that the encoded
file path was already normalised which means that normalising again is
just unnecessarily slow.

Furthermore, it means that Binary instances
in ghcide fail to roundtrip since we rely on empty file paths. We
could fix that on the ghcide side but it’s somewhat brittle and I
don’t see a reason to not address this here.


cocreature:
Actually this doesn’t solve the issue in ghcide since the conversion to a URI will fail on Windows. I still think this is preferable to normalising again but if you disagree feel free to close this.

PR #251, opened 2020-08-21
Add runWith for transport channels other than stdio

paulyoung:
This is a non-breaking change that allows for transport channels other than stdio.

I'm not sure about the inconsistency of strict vs lazy `ByteString` in the API (for input and output respectively) but it fell out of the existing types.

### DIFF
--- a/src/Language/Haskell/LSP/Control.hs
+++ b/src/Language/Haskell/LSP/Control.hs
@@ -8,6 +8,7 @@
 module Language.Haskell.LSP.Control
   (
     run
+  , runWith
   , runWithHandles
   ) where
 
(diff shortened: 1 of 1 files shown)


paulyoung:
Is there anything I can do to help move this along?

The build failures appear to be unrelated to my changes and are present on `master` and other recent PR builds. It looks like the macOS CI might be misconfigured.

wz1000:
Hi, we can certainly merge this now. However, we have been working on a pretty big overhaul of this library in https://github.com/alanz/haskell-lsp/pull/244, which is very close to landing. I'm curious, are you actually using this library. If so, I would like to volunteer to help patch your project to support the upcoming version.

paulyoung:
Thanks!

I'm using this in a traditional sense but I also required this patch so I could have a mode that uses a WebSocket for transport.

I appreciate the offer but my work is in a private repo for now. It's still fairly early so there might not be a lot to do though.

I followed the `reactor` approach in the example so can hopefully start with following however the example is updated.

You may be interested in the Dhall language server: https://github.com/dhall-lang/dhall-haskell/tree/master/dhall-lsp-server

PR #257, opened 2020-10-02
Prevent crash when log file is misconfigured

jhrcek:
This is my attemp at fixing https://github.com/haskell/haskell-language-server/issues/452
by falling back to stderr loggin in cases when the attempt to open the log file fails.

### DIFF
--- a/src/Language/Haskell/LSP/Core.hs
+++ b/src/Language/Haskell/LSP/Core.hs
@@ -1046,7 +1046,7 @@ setupLogger :: Maybe FilePath -> [String] -> Priority -> IO ()
 setupLogger mLogFile extraLogNames level = do
 
   logStream <- case mLogFile of
-    Just logFile -> openFile logFile AppendMode
+    Just logFile -> openFile logFile AppendMode `E.catch` handleIOException logFile
     Nothing      -> return stderr
   hSetEncoding logStream utf8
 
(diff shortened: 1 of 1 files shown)


Anrock:
IMO, expected behaviour would be to create log file. It's not like I manually create logfiles for all the software on my pc :thinking:

jhrcek:
Well, `openFile filePath AppendMode` DOES create the file, IF the directory which you provide exists.

But it
1. doesn't deal with stuff like `~` 
2. doesn't create directories on the path to file.

But if you think it's the right approach, I could extend this PR to [createDirectoryIfMissing](https://hackage.haskell.org/package/directory-1.3.6.1/docs/System-Directory.html#v:createDirectoryIfMissing) which fixes problem 2., but it doesn't help with 1.

bubba:
This makes sense to me, thanks! If its not too much trouble could you make this PR against wz1000:singleton-methods [here](https://github.com/alanz/haskell-lsp/pulls)? We're going to merge it soon and it would be nice to avoid the merge conflicts

Anrock:
@jhrcek sorry, my bad. I didn't realize that `AppendMode` also creates file if it doesn't exist.

jhrcek:
@bubba sure, opened here https://github.com/wz1000/haskell-lsp/pull/3

bubba:
Thanks, this should make it into the 1.0.0.0 release then

PR #258, opened 2020-10-12
Rename Language.Haskell.LSP to Language.LSP

bubba:
Making a PR for this to get people's opinions on this


Anrock:
Why tho

bubba:
@Anrock the package has been renamed from haskell-lsp to lsp, so we should drop the old nomenclature now and get the growing pains out of the way. Once we publish 1.0.0.0 it will be hard to justify breaking changes like this again. Pretty much everything requires manual migration anyway, renaming your imports will be the least of your worries!

paulyoung:
I think it makes it clearer that this package is for LSP support in other languages too, not just Haskell.

alanz (approved):

PR #261, opened 2020-10-14
Add GitHub Actions CI, fix 8.4.4 builds

bubba:
Given the discussion in https://github.com/haskell/haskell-language-server/issues/6#issuecomment-680646155, I thought it would make more sense to migrate to a new ci provider rather than fix the existing circleci config


wz1000:
Great, how do we disable circleci once and for all?

alanz:
I think I may have to do it, let me look

alanz:
I just did the "Stop Building" thing, which says

> Stop Building
>
> If you choose to stop building lsp on CircleCI, we will halt all ongoing builds, and any teammates who are currently following the project will automatically `unfollow` it.

PR #265, opened 2020-11-26
Add proper support for prepareRename

### LINKED ISSUE #264: (ReponseResult TextDocumentPrepareRename) needs to be nullable
Currently it is defined as:

```haskell
  ResponseResult TextDocumentPrepareRename     = Range |? RangeWithPlaceholder
```

However, in the specification it is:

```text
result: Range | { range: Range, placeholder: string } | null
```

This `null` is crucial, since the request is essentially a yes/no answer that indicated whether renaming at this location is allowed, and the “no” answer is being encoded as `null`.

Note that there is [this comment](https://github.com/alanz/lsp/blob/cedf0a49165c70ca0c4b1f92677e75d1fc129a17/lsp-types/src/Language/LSP/Types/Message.hs#L154-L157), which actually does not hold true in this case.

kirelagin:
There are two commits here.

The first commit fixes #264 by allowing a `null` response to `prepareRename`.

The second commit actually correctly sets `prepareProvider` in `RenameOptions` if the client has `prepareSupport` and the server provides a `TextDocumentPrepareRename` handler.

I checked that this works with VSCodium. If I try to rename a keyword and the server returns `Nothing` from the prepare handler, the UX reacts by showing a predefined error message, as expected:

image

### DIFF
--- a/lsp-types/src/Language/LSP/Types/Message.hs
+++ b/lsp-types/src/Language/LSP/Types/Message.hs
@@ -193,7 +193,7 @@ type family ResponseResult (m :: Method f Request) :: Type where
   ResponseResult TextDocumentOnTypeFormatting  = List TextEdit
   -- Rename
   ResponseResult TextDocumentRename            = WorkspaceEdit
-  ResponseResult TextDocumentPrepareRename     = Range |? RangeWithPlaceholder
+  ResponseResult TextDocumentPrepareRename     = Maybe (Range |? RangeWithPlaceholder)
   -- FoldingRange
   ResponseResult TextDocumentFoldingRange      = List FoldingRange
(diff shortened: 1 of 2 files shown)


bubba (approved):
Thanks, this makes sense to me!

bubba:
I'm noticing that the rename response should also be nullable, since the specification defines it as `WorkspaceEdit | null`. Presumably if it's null no renaming should occur, and like with this case it isn't an error.

kirelagin:
I believe, this is not actually needed, since the rename response is already “naturally nullable” in multiple ways: you can either set both `changes` and `documentChanges` to `Nothing`s or you can set either of them to `Just mempty`. I am not exactly sure how the clients are supposed to behave in each of the cases and adding yet another way to return an empty response certainly is not going to make the situation easier :).

PR #267, opened 2020-12-10
Add support for file and folder operations (create, rename, move) to workspace edits

banacorn:
Instead of having `TextDocumentEdit` in `WorkspaceEdit`

```haskell
data WorkspaceEdit =
  WorkspaceEdit
    { _changes         :: Maybe WorkspaceEditMap
    , _documentChanges :: Maybe (List TextDocumentEdit)
    } deriving (Show, Read, Eq)
```

It is now replaced by a new type called `DocumentChange`

```haskell
data WorkspaceEdit =
  WorkspaceEdit
    { _changes         :: Maybe WorkspaceEditMap
    , _documentChanges :: Maybe (List DocumentChange)
    } deriving (Show, Read, Eq)
```

Which is just a synonym of union of `WorkspaceEdit` and other operations

```haskell
type DocumentChange = TextDocumentEdit |? CreateFile |? RenameFile |? DeleteFile
```

So that it complies with the [spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspaceEdit) as suggested by @EggBaconAndSpam in https://github.com/alanz/lsp/issues/134#issuecomment-502025882

---

Here's how to construct a `DocumentChange` that creates a file for example: 
```haskell
InR (InL (CreateFile FileResourceChangeCreate "/path/to/new/file" Nothing))
```

But it's a bit wordy and unsafe because of the `FileResoureChangeKind`.
I'm considering hiding these constructors and export only smart constructors.

### DIFF
--- a/lsp-types/src/Language/LSP/Types/WorkspaceEdit.hs
+++ b/lsp-types/src/Language/LSP/Types/WorkspaceEdit.hs
@@ -58,7 +58,7 @@ deriveJSON lspOptions ''CreateFileOptions
 data CreateFile =
   CreateFile
     { -- | The resource to create.
-      _uri      :: Text
+      _uri      :: Uri
       -- | Additional options
     , _options  :: Maybe CreateFileOptions
     } deriving (Show, Read, Eq)
@@ -95,9 +95,9 @@ deriveJSON lspOptions ''RenameFileOptions
 data RenameFile =
   RenameFile
     { -- | The old (existing) location.
-      _oldUri   :: Text
+      _oldUri   :: Uri
       -- | The new location.
-    , _newUri   :: Text
+    , _newUri   :: Uri
       -- | Rename options.
     , _options  :: Maybe RenameFileOptions
     } deriving (Show, Read, Eq)
@@ -136,7 +136,7 @@ deriveJSON lspOptions ''DeleteFileOptions
 data DeleteFile =
   DeleteFile
     { -- | The file to delete.
-      _uri      :: Text
+      _uri      :: Uri
       -- | Delete options.
     , _options  :: Maybe DeleteFileOptions
     } deriving (Show, Read, Eq)
--- a/lsp-types/src/Language/LSP/Types/WorkspaceEdit.hs
+++ b/lsp-types/src/Language/LSP/Types/WorkspaceEdit.hs
@@ -38,6 +38,98 @@ deriveJSON lspOptions ''TextDocumentEdit
 
 -- ---------------------------------------------------------------------
 
+-- | For tagging `CreateFile`/`RenameFile`/`DeleteFile`
+-- Should this be merged with `ResourceOperationKind` ?
+data FileResourceChangeKind
+  = FileResourceChangeCreate
+  | FileResourceChangeRename
+  | FileResourceChangeDelete 
+  deriving (Read, Show, Eq)
+  
+instance ToJSON FileResourceChangeKind where
+  toJSON FileResourceChangeCreate = String "create"
+  toJSON FileResourceChangeRename = String "rename"
+  toJSON FileResourceChangeDelete = String "delete"
+
+instance FromJSON FileResourceChangeKind where
+  parseJSON (String "create") = pure FileResourceChangeCreate
+  parseJSON (String "rename") = pure FileResourceChangeRename
+  parseJSON (String "delete") = pure FileResourceChangeDelete
+  parseJSON _                 = mempty
+
+-- | Options to create a file.
+data CreateFileOptions =
+  CreateFileOptions
+    { -- | Overwrite existing file. Overwrite wins over `ignoreIfExists`
+      _overwrite      :: Bool
+      -- | Ignore if exists.
+    , _ignoreIfExists :: Bool
+    } deriving (Show, Read, Eq)
+
+deriveJSON lspOptions ''CreateFileOptions
+
+-- | Create file operation
+data CreateFile =
+  CreateFile
+    { _kind     :: FileResourceChangeKind
+      -- | The resource to create.
+    , _uri      :: Text
+      -- | Additional options
+    , _options  :: Maybe CreateFileOptions
+    } deriving (Show, Read, Eq)
+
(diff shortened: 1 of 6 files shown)


banacorn:
@bubba I should make a PR on [lsp-test](https://github.com/bubba/lsp-test), right? 
In that case, `lsp-test` would have to depend on this commit of `lsp` then 👀

bubba:
Hi @banacorn, thanks for this PR! Looks like there's a bit of a chicken and the egg situation with lap-types and lsp-test. You can create a branch on lsp-test for now and reference that commit in `cabal.project`, and then once this is merged we can merge the PRs for lsp-test

lsp-types/src/Language/LSP/Types/WorkspaceEdit.hs
```diff
@@ -38,12 +40,109 @@ deriveJSON lspOptions ''TextDocumentEdit
...
+  parseJSON (String "delete") = pure FileResourceChangeDelete
+  parseJSON _                 = mempty
+
+-- | Options to create a file.
+data CreateFileOptions =
+  CreateFileOptions
+    { -- | Overwrite existing file. Overwrite wins over `ignoreIfExists`
+      _overwrite      :: Bool
+      -- | Ignore if exists.
+    , _ignoreIfExists :: Bool
```
bubba, line 68: To match the specification these should be maybes:

```typescript
/**
 * Options to create a file.
 */
export interface CreateFileOptions {
	/**
	 * Overwrite existing file. Overwrite wins over `ignoreIfExists`
	 */
	overwrite?: boolean;
	/**
	 * Ignore if exists.
	 */
	ignoreIfExists?: boolean;
}
```

```suggestion
    { -- | Overwrite existing file. Overwrite wins over `ignoreIfExists`
      _overwrite      :: Maybe Bool
      -- | Ignore if exists.
    , _ignoreIfExists :: Maybe Bool
```

lsp-types/src/Language/LSP/Types/WorkspaceEdit.hs
```diff
@@ -38,12 +40,109 @@ deriveJSON lspOptions ''TextDocumentEdit
...
+
+deriveJSON lspOptions ''CreateFile
+
+-- Rename file options
+data RenameFileOptions =
+  RenameFileOptions
+    { -- | Overwrite target if existing. Overwrite wins over `ignoreIfExists`
+      _overwrite      :: Bool
+      -- | Ignores if target exists.
+    , _ignoreIfExists :: Bool
```
bubba, line 91:
```suggestion
      _overwrite      :: Maybe Bool
      -- | Ignores if target exists.
    , _ignoreIfExists :: Maybe Bool
```

lsp-types/src/Language/LSP/Types/WorkspaceEdit.hs
```diff
@@ -38,12 +40,109 @@ deriveJSON lspOptions ''TextDocumentEdit
...
+
+deriveJSON lspOptions ''RenameFile
+
+-- Delete file options
+data DeleteFileOptions =
+  DeleteFileOptions
+    { -- | Delete the content recursively if a folder is denoted.
+      _recursive          :: Bool
+      -- | Ignore the operation if the file doesn't exist.
+    , _ignoreIfNotExists  :: Bool
```
bubba, line 116:
```suggestion
      _recursive          :: Maybe Bool
      -- | Ignore the operation if the file doesn't exist.
    , _ignoreIfNotExists  :: Maybe Bool
```

lsp-types/src/Language/LSP/Types/WorkspaceEdit.hs
```diff
@@ -38,12 +40,109 @@ deriveJSON lspOptions ''TextDocumentEdit
...
+instance ToJSON FileResourceChangeKind where
+  toJSON FileResourceChangeCreate = String "create"
+  toJSON FileResourceChangeRename = String "rename"
+  toJSON FileResourceChangeDelete = String "delete"
+
+instance FromJSON FileResourceChangeKind where
+  parseJSON (String "create") = pure FileResourceChangeCreate
+  parseJSON (String "rename") = pure FileResourceChangeRename
+  parseJSON (String "delete") = pure FileResourceChangeDelete
+  parseJSON _                 = mempty
```
bubba, line 60: Bit of a nitpick here, but I think we can avoid having to create this data type if we omit storing the `kind` in `{Create,Rename,Delete}File`. Instead we can just parse it and emit the `kind` field in custom `ToJSON` instances, like how `WorkDoneProgressBegin` is handled

https://github.com/alanz/lsp/blob/8d6e8f316a5a1c40986cdc6400611376c8ab0140/lsp-types/src/Language/LSP/Types/Progress.hs#L89-L107

I think this more closely matches the string literal tags in TypeScript interfaces, and it means users can't accidentally create a dodgy `CreateFile FileResourceChangeRename x`

bubba (commented):
Thanks, this is great! Just some minor nitpicks about the string literal tag. Out of curiosity, are you using this for your own language server?

lsp-types/src/Language/LSP/Types/WorkspaceEdit.hs
banacorn: oopsy!

lsp-types/src/Language/LSP/Types/WorkspaceEdit.hs
banacorn: Nice!

banacorn:
Yes, I'm building a language server for my toy language. 
I have a button that exports all proof obligations to a new file, that's why I need this feature.

I'm planning on building a language server for Agda next.

bubba:
Whoops, looks like fix keyword in the PR message closes PRs across GitHub accounts

bubba (approved):
LGTM now. Let me know when you're happy/referencing the new lsp-test commit and I'll merge

bubba:
Looks like its passing now. Want me to merge @banacorn ?

banacorn:
Yes, please! Thank you!

Issue #263, opened 2020-11-09
 Could not deduce instance of GEq, dependency version unspecified


banacorn:
A side issue from https://github.com/alanz/lsp/pull/262

`lsp-types` doesn't build with `stack-8.6.5.yaml` and `stack-8.6.4.yaml`, but `stack-8.8.2.yaml` and `stack-8.8.1.yaml` work fine on the other hand.

I think some of the dependency version lower bound need to be specified, but I'm not sure which one it is.

```
 /Users/banacorn/haskell/lsp/lsp-types/src/Language/LSP/Types/Message.hs:620:13: error:
     • Could not deduce (GEq SMethod) arising from a use of ‘geq’
       from the context: (t1 ~ 'Notification, m1 ~~ m,
                          HasJSON (ServerMessage m), Message m ~ NotificationMessage m)
         bound by a pattern with constructor:
                    IsServerNot :: forall (m :: Method 'FromServer 'Notification).
                                   (HasJSON (ServerMessage m), Message m ~ NotificationMessage m) =>
                                   ServerNotOrReq m,
                  in a case alternative
         at src/Language/LSP/Types/Message.hs:619:4-14
       or from: (t ~ 'Notification, m2 ~~ m4, HasJSON (ServerMessage m4),
                 Message m4 ~ NotificationMessage m4)
         bound by a pattern with constructor:
                    IsServerNot :: forall (m :: Method 'FromServer 'Notification).
                                   (HasJSON (ServerMessage m), Message m ~ NotificationMessage m) =>
                                   ServerNotOrReq m,
                  in a case alternative
         at src/Language/LSP/Types/Message.hs:619:17-27
     • In a stmt of a 'do' block: Refl <- geq m1 m2
       In the expression:
         do Refl <- geq m1 m2
            pure HRefl
       In a case alternative:
           (IsServerNot, IsServerNot)
             -> do Refl <- geq m1 m2
                   pure HRefl
     |
 620 |     Refl <- geq m1 m2
     |             ^^^^^^^^^
```

<details><summary>here are the all 4 error messages (including the one above)</summary>

```
 /Users/banacorn/haskell/lsp/lsp-types/src/Language/LSP/Types/Message.hs:620:13: error:
     • Could not deduce (GEq SMethod) arising from a use of ‘geq’
       from the context: (t1 ~ 'Notification, m1 ~~ m,
                          HasJSON (ServerMessage m), Message m ~ NotificationMessage m)
         bound by a pattern with constructor:
[...]

PR #273, opened 2021-01-28
Fix dependency problems in #263

banacorn:
This is a follow up PR to https://github.com/alanz/lsp/issues/263#issuecomment-764777921.

Stack now builds with `stack-8.8.2.yaml`, `stack-8.8.1.yaml`, `stack-8.6.5.yaml`, `stack-8.6.4.yaml`.

However, I cannot resolve dependency deadlocks in `stack-8.4.4.yaml` and below. 
Should we drop support for `stack-8.4.4.yaml` and below and remove those `.yaml` files?


banacorn:
I'm marking it "Ready for review" just to trigger the CI.
This is still a work in progress, please don't merge it yet.

banacorn:
> I'm marking it "Ready for review" just to trigger the CI.

Actually I'm not sure if this is necessary anymore, and GitHub action is currently kaput anyway 😄  https://www.githubstatus.com/incidents/tf9v5jjmq2lg

banacorn:
I'm ready to merge!

wz1000:
The stack.yamls look good, could you drop the bounds in the cabal file? https://github.com/alanz/lsp/pull/274 fixes the root cause of the problem in a laxer way.

bubba:
Seems like cabal can build fine on 8.4.4, looking into those weird stack-8.4.4.yaml errors

bubba:
I couldn't push to your branch so I've created another PR with the stack 8.4 fixes in #276

alanz:
> The stack.yamls look good, could you drop the bounds in the cabal file? #274 fixes the root cause of the problem with looser constraints.

Building master with cabal/ghc-8.10.3 still has constraint problems.

wz1000:
What is the error you are getting?

alanz:
I get the same error s #263

alanz:
@wz1000 apologies, the problem *is* gone. I was building it as a dependency of my project, rather than directly.

PR #277, opened 2021-02-03
use Text instead of String

wz1000:


bubba (commented):
Should we also get
https://github.com/alanz/lsp/blob/793810cb1681e293f7b5fbdac6821e51e87e3c17/lsp-types/src/Language/LSP/Types/DocumentLink.hs#L64-L66

bubba (approved):

wz1000:
> Should we also get
> 
> https://github.com/alanz/lsp/blob/793810cb1681e293f7b5fbdac6821e51e87e3c17/lsp-types/src/Language/LSP/Types/DocumentLink.hs#L64-L66

Fixed.

bubba:
Argh, we can ignore the func-tests for now

wz1000:
What do you think about merging this repo and lsp-test? I think it makes sense now that lsp-test is used to test lsp

/cc @alanz - the problem is that changing a type in lsp-types means lsp-test no longer compiles, so the test suite fails - the changes have to be co-ordinated across the repos

bubba:
It's been on the back of my mind but yeah I think it would make a lot of sense to merge them at some point.

alanz:
I agree they could/should be merged. And the ghcide merge did preserve history, issues etc, with some work.

If we did do that, we should perhaps move it into haskell org at the same time. Otherwise the one owner will be weird, whatever we decide.

bubba:
Is Jasper still the right person to contact about this?

alanz:
I would talk to him as the first option.

Alan

PR #281, opened 2021-02-10
Add support for ghc-9.0.1

jneira:
* As first attempt, listing upstream packages needing patches in `allow-newer`
* This only can be built with cabal-3.4.0.0


jneira:
Closed in favor of #312

PR #284, opened 2021-02-18
Prevent crashing when optional fields are missing

anka-213:
More specifically, when VS Code sent an "exit" message it was missing the "params" field, which should be fine, 
 since it is optional and a "null" is expected, but missing is not treated the same as null in this case.

This solution is kind of a hack and only solves this specific case, but it works.

See also this issue: https://github.com/haskell/aeson/issues/646

### DIFF
--- a/lsp-types/src/Language/LSP/Types/Message.hs
+++ b/lsp-types/src/Language/LSP/Types/Message.hs
@@ -253,8 +253,15 @@ deriving instance Eq   (MessageParams m) => Eq (RequestMessage m)
 deriving instance (Read (SMethod m), Read (MessageParams m)) => Read (RequestMessage m)
 deriving instance Show (MessageParams m) => Show (RequestMessage m)
 
+-- | Replace a missing field in an object with a null field, to simplify parsing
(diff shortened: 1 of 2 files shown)


wz1000:
How about `Maybe Void`? /cc @bubba

anka-213:
@wz1000 As far as I can tell (haven't tested) that wouldn't work, since the `genericParseJSON` doesn't handle `Maybe` in a type parameter (see this issue https://github.com/haskell/aeson/issues/690). Here we are using type families to determine the type, which certainly won't work with the blanket instance we have.

tolysz:
How do I test it end-to-end? I would like to test it with the modified `aeson`...

anka-213:
@tolysz This json string should do it:
```
{
  "jsonrpc": "2.0",
  "id": 15,
  "method": "shutdown"
}
```

I'm not completely sure how to actually plug it in though.

tolysz:
Does this project have some tests? Could it be added there and failing?
So I could just do my changes and run `stack test` ?

anka-213:
@tolysz Now I've added a test case that fails if you revert my fix.

wz1000:
Thanks

PR #287, opened 2021-02-25
Add simple benchmark

wz1000:


jneira (approved):
looks great and it already has been useful

PR #288, opened 2021-02-25
Split out parsing into its own module and add a few inline pragmas

wz1000:
This leads to a ~5-10% improvement in allocations/time when measured against
the benchmark added in https://github.com/alanz/lsp/pull/287

Also changed the heterogeneous equality functions to compare methods to handle custom methods.


jneira (approved):
lgtm

PR #289, opened 2021-02-25
Use StrictData for lsp-types

wz1000:
Also added a few strictness annotations and replaced the use of
`Data.Map` with `Data.Map.Strict`

Removed a few depreceated warnings of some fields because it is impossible to
create those records without triggering the warnings.


bubba:
This is interesting, have we ran any benchmarks with this on `ghcide/hls` with strictdata on?

PR #291, opened 2021-02-26
Add some missing Completion-related features

michaelpj:
- `InsertReplaceTextEdit`
    - Needs a new type, I put it in the `Completion` module since it's grouped with those types in the spec.
- `InsertMode`
    - Straightforward enum
- `resolveSupport`
    - It looks like we already have methods for this, we just didn't have the client capability listed.


jneira (approved):
LGTM, many thanks for your recent work here, spec compliance was a little bit abandoned

bubba:
Thanks again!

PR #292, opened 2021-02-26
Fix applying a TextEdit past the end of the document

### LINKED ISSUE #271: Incorrect positioning of text in applyTextEdit
Presumably, a text insertion positioned beyond the end of the document should be appended to it (or, possibly more logically, enough white space and line returns should be added to locate the inserted text in the required position).

However, in Language.LSP.Types.WorkspaceEdit.applyTextEdit, it seems that the insert line is set to the maximum between the requested line and the max number of lines, but the column number is not changed.

Example, Position 20 3 is interpreted as Position 1 3:

```
>>> applyTextEdit (TextEdit (Range (Position 1 3) (Position 1 3)) "ZZZ") "aaaaa\nbbbbbb"
"aaaaa\nbbbZZZbbb"

>>> applyTextEdit (TextEdit (Range (Position 20 3) (Position 20 3)) "ZZZ") "aaaaa\nbbbbbb"
"aaaaa\nbbbZZZbbb"
```

michaelpj:
A simple fix, the text gets appended. You could imagine being smarter
here, but at least it doesn't insert it into the existing text as it
would before.

Fixes #271.

### DIFF
--- a/lsp-types/src/Language/LSP/Types/WorkspaceEdit.hs
+++ b/lsp-types/src/Language/LSP/Types/WorkspaceEdit.hs
@@ -274,6 +274,9 @@ applyTextEdit (TextEdit (Range sp ep) newText) oldText =
   where
     splitAtPos :: Position -> Text -> (Text, Text)
     splitAtPos (Position sl sc) t =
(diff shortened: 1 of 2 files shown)


bubba:
Thanks!

PR #298, opened 2021-03-05
Fix lsp-test example

michaelpj:
I noticed this was broken doing `cabal build all`.

- Fix compilation of the test.
- Depend on and use `lsp-demo-reactor-server` instead of HLS so it is
actually testable in isolation.
- Teach `lsp-demo-reactor-server` to respond to `TextDocumentSymbol`
(could have changed the example but this was easy enough).
- Ensure that the CI builds and tests it so we won't miss it in future.


wz1000:
Thanks

PR #300, opened 2021-03-06
Organize repository into three top-level folders for the packages

michaelpj:
Implements the suggestions in https://github.com/haskell/lsp/issues/299 and also merges or removes some redundant or repeated files.


bubba (approved):
Thanks, this makes a lot of sense to me. Happy to merge if you are @wz1000

wz1000:
@michaelpj there are a couple of conflicts now

PR #301, opened 2021-03-06
Add tag support for DocumentSymbol, SymbolInformation, and document symbol provider label

michaelpj:
Another couple of 3.16 things.


michaelpj:
Also WorkspaceSymbol needed it.

PR #312, opened 2021-03-30
Add support for ghc 9.0.1

anka-213:
This integrates the changes from #281 and also fixes some Template Haskell issues caused by [ghc-9.0.1 having stricter staging rules than earlier versions](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.0#the-order-of-th-splices-is-more-important).

(Step towards https://github.com/haskell/haskell-language-server/issues/297)

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -20,7 +20,7 @@ jobs:
     - uses: haskell/actions/setup@v1
       with:
         ghc-version: ${{ matrix.ghc }}
-        cabal-version: '3.2'
+        cabal-version: '3.4'
 
     - name: Cache
       uses: actions/cache@v1
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -34,6 +34,8 @@ jobs:
           ${{ runner.os }}-build-
           ${{ runner.os }}-
 
+    - name: Cabal update
+      run: cabal update
     - name: Build
       run: cabal build all
     - name: Test
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc: ['8.10.4', '8.8.3', '8.6.5', '8.4.4']
+        ghc: ['9.0.1', '8.10.4', '8.8.3', '8.6.5', '8.4.4']
         os: [ubuntu-latest, macOS-latest, windows-latest]
         exclude:
           - os: windows-latest
--- a/cabal.project
+++ b/cabal.project
@@ -18,22 +18,10 @@ max-backjumps: 10000
 
 allow-newer:
   lens:template-haskell,
-  parallel:base,
-  hashable:base,
-  async:base,
-  quickcheck-instances:base,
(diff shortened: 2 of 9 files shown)


jneira:
thanks again, maybe some of libs in the allow-newer list has been updated and could be removed

anka-213:
@jneira Probably. Is there a way to automate that? My attempts requires me to loop between running `cabal build` and adding the single package it mentions to the list.
Can I tell cabal to display all packages that requires `allow-newer`?

anka-213:
I've added references to https://github.com/mokus0/th-extras/pull/8 and https://github.com/obsidiansystems/dependent-sum/pull/57 in the `cabal.project`, so this can be built even before they are merged. They should be removed once those pull requests are merged.

anka-213:
I'm not sure why it failed on CI. It worked for me locally. :/

jneira:
That is my manual process too, I don't know an automate way to do it 😟

anka-213:
@jneira  If you have time, can you check if this patch works for you on ghc-9.0.1 or if it accidentally works for me locally because of some impurity?

jneira:
@anka-213 I cant reproduce the error after a cabal update (sadly i didnt try before) 
I've observed github ci has no `cabal update` step so maybe the haskell action left it in a old index state, i would try to add a cabal update before start the build.

anka-213:
@jneira Yep, that was the problem. Thanks!

jneira:
@anka-213 hi! i think we could merge this one and remove one of the remote repo packages.
could you rebase master?, i hope it will not take much time.

jneira:
@anka-213 i tried to fix the build with e96383a

anka-213:
Those were introduced in this commit: https://github.com/haskell/lsp/commit/ef59c28b41ed4c5775f0ab0c1e985839359cec96#diff-b8ed06757045fac949c89f2139a862498ad2b6d1f82c61a31e7c91d6cf0eaa70

Is that problem resolved now?

jneira:
he I suggested the change to fix the build there but this pr seems to fix it in another way, to investigate

anka-213:
Oh, the answer is really simple. My changes includes upstream patches of those packages, so we no longer need to constrain the version.

jneira:
the question is we can't do a hackage release until all upstream packages are there so not sure if we should merge this
the changes are not too pervasive so maybe there won't be much conflicts (and there is no too much prs to merge in master), thoughts?

anka-213:
Oh, right. That's a good point. I didn't think of that. Otoh, it does still build successfully (I believe, haven't actually tried) on any ghc < 9 without those upstream patches, so it shouldn't really be a problem to put it on hackage.

pepeiborra:
Are you sure that a Hackage release cannot be done? Technically, the changes to `cabal.project` are not impeding a Hackage release, as long as the package builds with a vanilla `cabal.project`.

jneira:
@pepeiborra mmm you are right, it will not work for 9.0.1, failing in the solver phase like the actual one (i've checked it, just in case it would be a build error and no solver error)

jneira (approved):
lgtm @anka-213 many thanks for working on this
will merge soon if nobody disagree

PR #315, opened 2021-04-03
Add exception instances to LspT

pepeiborra:


wz1000:
I guess these don't hurt, but wouldn't it be best to use `unliftio` for this? Then you would have some amount of async exception safety as well.

pepeiborra:
I'm not familiar with `unliftio`, how is it async exception safe?

wz1000:
[UnliftIO.Exception](https://hackage.haskell.org/package/unliftio-0.2.13/docs/UnliftIO-Exception.html) provides all the traditional exception handing functions (`try`,`mask`,`catch`...), but these functions are based on the corresponding versions from [`safe-exceptions`](https://hackage.haskell.org/package/safe-exceptions) which treat async exceptions properly, by attempting to detect them deal with them appropriately.

pepeiborra:
I'm not sure how to implement [`generalBracket`](https://hackage.haskell.org/package/exceptions-0.10.4/docs/Control-Monad-Catch.html#v:generalBracket) with unliftio. 

Would a `MonadCatch` instance that uses unliftio + derived `MonadThrow` & `MonadMask` instances make any sense?

wz1000:
Does [`bracketOnError`](https://hackage.haskell.org/package/unliftio-0.2.13/docs/UnliftIO-Exception.html#v:bracketOnError) work?

pepeiborra:
I don't think so, with `generalBracket` the release handler owns the choice, so I would need to run the release handler in order to know whether to use `bracket` or `bracketOnError`

PR #316, opened 2021-04-06
Add nix

berberman:


jneira (commented):
Thanks for adding it, not nix here though

jneira:
@berberman thanks! i am gonna merging it, not sure if it would need some changes, but it can be done in a new pr

PR #317, opened 2021-04-06
lsp-test: catch IO exception when sending messages to server

berberman:


jneira (approved):
LGTM, thanks for this small but useful change

PR #320, opened 2021-04-13
Improve error messages on JSON decode failures

strager:
If we receive a JSON payload from the LSP server which has unexpected
values for some fields, 'Aeson.parseMaybe' fails and the program crashes
because of 'fromJust'. Make two improvements:

* Use 'Aeson.parse' instead to get an error message
* In 'Aeson.FromJSON' instances, avoid 'mempty'; use 'fail'
  to give a more helpful message instead. (I would have used
  'Aeson.parseFail', but older versions of Aeson don't have that
  function.)

Example before:

    myprogram: Maybe.fromJust: Nothing
    CallStack (from HasCallStack):
      error, called at libraries/base/Data/Maybe.hs:148:21 in base:Data.Maybe
      fromJust, called at src/Language/LSP/Test/Decoding.hs:86:42 in lsp-test-0.14.0.0-6s4OGLFS3RcJa58gpuLS2v:Language.LSP.Test.Decoding

Example after:

    myprogram: expected 'null'
    CallStack (from HasCallStack):
      error, called at src/Language/LSP/Test/Decoding.hs:95:25 in lsp-test-0.14.0.0-6s4OGLFS3RcJa58gpuLS2v:Language.LSP.Test.Decoding

### DIFF
--- a/lsp-test/src/Language/LSP/Test/Decoding.hs
+++ b/lsp-test/src/Language/LSP/Test/Decoding.hs
@@ -83,15 +83,16 @@ getRequestMap = foldl' helper emptyIxMap
     _ -> acc
 
 decodeFromServerMsg :: RequestMap -> B.ByteString -> (RequestMap, FromServerMessage)
-decodeFromServerMsg reqMap bytes = unP $ fromJust $ parseMaybe p obj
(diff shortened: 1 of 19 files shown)


wz1000:
Thanks, looks good.

PR #321, opened 2021-04-13
Do not crash on workspace/didChangeConfiguration

strager:
The eslint-server LSP server sends the following message to the client:

    {
        "jsonrpc": "2.0",
        "params": {
            "registrations": [
                {
                    "registerOptions": {},
                    "method": "workspace/didChangeConfiguration",
                    "id": "4a56f5ca-7188-4f4c-a366-652d6f9d63aa"
                }
            ]
        },
        "method": "client/registerCapability",
        "id": 0
    }

This crashes lsp-test because it asserts that registerOptions is null
(but an empty object is not null).

Make lsp-test accept either null or an object.

Warning: This commit affects every use of Empty, not just
WorkspaceDidChangeConfiguration.

### DIFF
--- a/lsp-types/src/Language/LSP/Types/Common.hs
+++ b/lsp-types/src/Language/LSP/Types/Common.hs
@@ -9,6 +9,7 @@ module Language.LSP.Types.Common where
 import Control.Applicative
 import Control.DeepSeq
 import Data.Aeson
+import qualified Data.HashMap.Strict as HashMap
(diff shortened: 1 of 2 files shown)


wz1000:
> What's the right way to handle this?

Something like this could work: https://github.com/haskell/lsp/pull/284

> Where do we write tests for this change?

There are some tests in `lsp/tests`

PR #331, opened 2021-05-11
Add test files to lsp-test extra-source-files

bubba:

PR #327, opened 2021-04-23
Fix and revamp Stack dependencies

banacorn:
1. Fix dependency problems (mostly related to `resourcet` from `conduit` in `lsp-test`)
2. Remove extraneous stack.yaml files
3. Add and keep those stack.yaml files that should be in the CI:
    * ghc-8.4.4 - lts-12.26
    * ghc-8.6.5 - lts-14.27
    * ghc-8.8.3 - lts-16.11
    * ghc-8.10.4 - lts-17.9

I'm going to make another PR to include Stack in the CI.
So that we can solve these Stack dependency problems once and for all.

----

edit: Stack image version changes made by this PR

| HLS    | LSP (current) | This PR |
| :----- | :------------ | :------: | 
|        | 8.4.2         |  removed  | 
|        | 8.4.3         |  removed  | 
|        | 8.4.4         |  kept     | 
| 8.6.4  | 8.6.4         |  kept  |
| 8.6.5  | 8.6.5         |  kept     |
|        | 8.8.1         |  removed  | 
| 8.8.2  | 8.8.2         |  removed  | 
| 8.8.3  |              | added |
| 8.8.4  |               | added |
| 8.10.2 |               | added |
| 8.10.3 |               | added |
| 8.10.4 |               | added |

PR #332, opened 2021-05-27
Call hierarchy support

July541:
As part of my GSoC 2021 works, this pr adds call hierarchy supporting according to [lsp spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_prepareCallHierarchy).

### DIFF
--- a/lsp-test/test/DummyServer.hs
+++ b/lsp-test/test/DummyServer.hs
@@ -202,4 +202,14 @@ handlers =
         if x == 0 && y == 0
           then resp $ Right Nothing
           else resp $ Right $ Just $ List [item]
+     , requestHandler SCallHierarchyIncomingCalls $ \req resp -> do
+        let RequestMessage _ _ _ params = req
+            CallHierarchyIncomingCallsParams _ _ item = params
+        resp $ Right $ Just $
+          List [CallHierarchyIncomingCall item (List [Range (Position 2 3) (Position 4 5)])]
+     , requestHandler SCallHierarchyOutgoingCalls $ \req resp -> do
+        let RequestMessage _ _ _ params = req
+            CallHierarchyOutgoingCallsParams _ _ item = params
+        resp $ Right $ Just $
+          List [CallHierarchyOutgoingCall item (List [Range (Position 4 5) (Position 2 3)])]
     ]
--- a/lsp-test/test/DummyServer.hs
+++ b/lsp-test/test/DummyServer.hs
@@ -16,12 +16,12 @@ import System.Directory
 import System.FilePath
 import System.Process
 import Language.LSP.Types
-  
+
 withDummyServer :: ((Handle, Handle) -> IO ()) -> IO ()
 withDummyServer f = do
   (hinRead, hinWrite) <- createPipe
   (houtRead, houtWrite) <- createPipe
-  
+
   handlerEnv <- HandlerEnv <$> newEmptyMVar <*> newEmptyMVar
   let definition = ServerDefinition
         { doInitialize = \env _req -> pure $ Right env
@@ -185,4 +185,21 @@ handlers =
                 Nothing
                 Nothing
         resp $ Right $ InR res
+     , requestHandler STextDocumentPrepareCallHierarchy $ \req resp -> do
+        let RequestMessage _ _ _ params = req
+            CallHierarchyPrepareParams _ pos _ = params
+            Position x y = pos
+            item =
+              CallHierarchyItem
+                "foo"
+                SkMethod
+                Nothing
+                Nothing
+                (Uri "")
+                (Range (Position 2 3) (Position 4 5))
+                (Range (Position 2 3) (Position 4 5))
+                Nothing
+        if x == 0 && y == 0
+          then resp $ Right Nothing
+          else resp $ Right $ Just $ List [item]
     ]
--- a/lsp-types/src/Language/LSP/Types/CallHierarchy.hs
+++ b/lsp-types/src/Language/LSP/Types/CallHierarchy.hs
@@ -1,5 +1,5 @@
-{-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE TemplateHaskell #-}
 
 {- | Since LSP 3.16.0 -}
 module Language.LSP.Types.CallHierarchy where
@@ -8,13 +8,13 @@ import Data.Aeson.TH
 import Data.Aeson.Types ( Value )
 import Data.Text ( Text )
 
-import Language.LSP.Types.Progress
-import Language.LSP.Types.TextDocument
-import Language.LSP.Types.StaticRegistrationOptions
-import Language.LSP.Types.DocumentSymbol
-import Language.LSP.Types.Uri
-import Language.LSP.Types.Location
 import Language.LSP.Types.Common
+import Language.LSP.Types.DocumentSymbol
+import Language.LSP.Types.Location
+import Language.LSP.Types.Progress
+import Language.LSP.Types.StaticRegistrationOptions
+import Language.LSP.Types.TextDocument
+import Language.LSP.Types.Uri
 import Language.LSP.Types.Utils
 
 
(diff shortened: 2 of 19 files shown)


jneira:
@July541 great work, i'll review it asap
//cc @michaelpj

lsp-types/src/Language/LSP/Types/CallHierarchy.hs
```diff
@@ -0,0 +1,100 @@
...
+  []
+deriveJSON lspOptions ''CallHierarchyRegistrartionOptions
+
+makeExtendingDatatype "CallHierarchyPrepareParams"
+  [''TextDocumentPositionParams, ''WorkDoneProgressParams] []
+deriveJSON lspOptions ''CallHierarchyPrepareParams
+
+data CallHierarchyItem =
+  CallHierarchyItem
+    { _name :: Text
```
michaelpj, line 44: Can we include haddock corresponding to the doc in the LSP spec for each item? Tedious, but matching the spec is a simple policy and more field documentation is almost always better.

lsp-types/src/Language/LSP/Types/CallHierarchy.hs
```diff
@@ -0,0 +1,100 @@
...
+data CallHierarchyClientCapabilities =
+  CallHierarchyClientCapabilities
+    { _dynamicRegistration :: Maybe Bool }
+    deriving (Show, Read, Eq)
+deriveJSON lspOptions ''CallHierarchyClientCapabilities
+
+makeExtendingDatatype "CallHierarchyOptions" [''WorkDoneProgressOptions] []
+deriveJSON lspOptions ''CallHierarchyOptions
+
+makeExtendingDatatype "CallHierarchyRegistrartionOptions"
```
michaelpj, line 30:
```suggestion
makeExtendingDatatype "CallHierarchyRegistrationOptions"
```

lsp-test/test/DummyServer.hs
```diff
@@ -185,4 +185,21 @@ handlers =
                 Nothing
                 Nothing
         resp $ Right $ InR res
+     , requestHandler STextDocumentPrepareCallHierarchy $ \req resp -> do
```
michaelpj, line 188: Should we test the other methods too?

michaelpj (approved):
LGTM, although I'm never sure all the places that need to be updated with capabilities etc!

lsp-types/src/Language/LSP/Types/CallHierarchy.hs
July541:
```haskell
-- <https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_prepareCallHierarchy>
data CallHierarchyItem = ...
```

Should it look like this? I have little knowledge about haddock.

lsp-types/src/Language/LSP/Types/CallHierarchy.hs
July541: My fault, thanks.

lsp-test/test/DummyServer.hs
July541: We should, I jumped to hls development and missed the other two, although it works well.

July541:
> although I'm never sure all the places that need to be updated with capabilities etc!

It turns out I updated the capabilities following the compile errors/warnings.

banacorn:
This looks great! 
Actually I have a [branch](https://github.com/banacorn/lsp/tree/call-hierarchy) with call hierarchy implemented, and I planned to make a PR after #327 is accepted ... It'd be nice it someone can merge it sooner ...

July541:
@banacorn I am sooo sorry to see about that, I checked the issue list and find no one explicitly intends to do this. To avoid the recurrence of such incidents, I add a [comment](https://github.com/haskell/haskell-language-server/issues/738#issuecomment-850125393) to make what I'm going to do more public in this repo.

jneira:
Ouch, that is unfortunate indeed, i was not aware of @banacorn plan neither :worried:

July541:
@banacorn I added some missing tests after compared both solutions of you and me, thank you for your inspiration:)

banacorn:
No, it's not your fault !!
I should've explicitly stated my intent and commented at that issue just like what you've done.
No one on earth would've known what I'm brewing in a secluded branch of my fork.

Anyway, my branch doesn't work. I was actually gonna make a PR and ask for help.
I hit some problem like in https://github.com/haskell/lsp/pull/314, none of my call hierarchy handlers are responding to the editor.

jneira:
Ok, what do you think about:
* @banacorn could review in deep this pr and propose changes (even including code changes that can be commited in the web ui)
* As @banacorn's branch already has been useful for this one, we would set @banacorn as co-author in the merge commit

July541:
> * As @banacorn's branch already has been useful for this one, we would set @banacorn as co-author in the merge commit

I agree with this.

@banacorn Out of curiosity, what do you mean about no call hierarchy response? I can navigate the call hierarchy on my client(VSCode) without any explicit configuration.

banacorn:
@July541 that means it's just broken, as if no call hierarchy supported has been added. 
I think I'll get to know what's wrong with my code after reviewing your PR. 🙂

bubba (approved):
Hi, I'm not too active here anymore but just chiming in here to say this looks great! Thanks for this PR and I hope you have a great summer of code :)

July541:
It seems #331 has the same confliction problem that failed on Windows with 8.10.4, I don't have any idea about it for the time being:(

jneira:
will try to take a look in my windows laptop

jneira:
it seems to be a dependency solver issue, related with dependent-map and some, maybe we would need to add some constraints in cabal.project similar to the stack.yaml of his itself

jneira:
I've got to build it adding

```
constraints: some == 1.0.1,
             dependent-sum == 0.7.1.0

max-backjumps: 10000
```

But i dont like the workaround

banacorn (changes requested):

lsp-types/src/Language/LSP/Types/Message.hs
```diff
@@ -193,6 +198,10 @@ type family ResponseResult (m :: Method f Request) :: Type where
   -- FoldingRange
   ResponseResult TextDocumentFoldingRange      = List FoldingRange
   ResponseResult TextDocumentSelectionRange    = List SelectionRange
+  -- Call hierarchy
+  ResponseResult TextDocumentPrepareCallHierarchy = Maybe (List CallHierarchyItem)
+  ResponseResult CallHierarchyIncomingCalls    = Maybe (List CallHierarchyIncomingCall)
+  ResponseResult CallHierarchyOutgoingCalls    = Maybe (List CallHierarchyOutgoingCall)
```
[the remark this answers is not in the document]
July541: I just keep it to satisfy the spec. And I still notice that we have several dealings about `null`, for example:
1. Use `Maybe` to represent `null`. https://github.com/haskell/lsp/blob/3e688d99218889cc17aaf43337ad9902aed321f5/lsp-types/src/Language/LSP/Types/Message.hs#L168
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_hover
2. Ignore `null`.
https://github.com/haskell/lsp/blob/3e688d99218889cc17aaf43337ad9902aed321f5/lsp-types/src/Language/LSP/Types/Message.hs#L169
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_signatureHelp
3. Strictly follow the spec like me.
https://github.com/haskell/lsp/blob/3e688d99218889cc17aaf43337ad9902aed321f5/lsp-types/src/Language/LSP/Types/Message.hs#L205
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_workspaceFolders

Maybe we need to unify them under one specific rule.

jneira:
> But i dont like the workaround

I dont like it but maybe can be applied to make progress until we have a proper solution

banacorn (approved):

jneira:
Merged! many thanks to @July541 and @banacorn too of course, amazing work

anka-213:
Stupid question, but what is a Call Hierarchy? I couldn't figure it out from the spec, which just seemed to describe the protocol for it, not what the actual thing is.

PR #334, opened 2021-06-10
Update link in lsp package description

jhrcek:
... to get more relevant link at the top of this page: https://hackage.haskell.org/package/lsp


jneira:
Many thanks for correct the info

PR #337, opened 2021-06-20
pass the lspConfig at initialization time as well

pepeiborra:
This is needed in the ghcide benchmark suite because some of the Config options are not dynamic (`checkParents` is the one I am trying to change) and need to be given at initialisation


jneira (approved):
lgtm

PR #338, opened 2021-06-20
Delete and gitignore hie.yaml

### LINKED ISSUE #336: How to get haskell-language-server to work _on_ lsp codebase?
I'm trying to get code navigation/auto completion working with `haskell/lsp` but it does not work. I'm using the stack workflow.

I complied lsp with  `stack build --stack-yaml stack-8.10.4` and it went without a hitch

Now, when I try to run `haskell-language-server` in the lsp directory I get 

```
...
...
Completed (0 files worked, 89 files failed)
```

Investigating further I find that there is a `hie.yaml` but that has cradle setting for `cabal` only. I'm much more comfortable with stack and rather not get into a cabal workflow.

How do I get things working with haskell-language-server + stack _on_ the lsp codebase?

michaelpj:
Implicit setup is good now, we don't need this, and people can write
their own one locally if they need to manually select a cabal or stack
cradle.

Also bonus gitignore `.envrc`.

Fixes #336.


jneira (approved):
lgtm, thanks!

PR #344, opened 2021-07-03
Rollback NFP interning

pepeiborra:
In retrospect I don't think interning NFPs is a good idea. It can lead to space leaks, the efficiency gains are not that huge, and in some cases we can achieve the same efficiencies by carefully tweaking the call sites to maximise sharing.


wz1000:
Alright. Do you have any benchmarks? How are you checking these claims? Would something like an LRU cache help fix the space leaks?

pepeiborra:
I am using telemetry metrics after running a patched version of Sigma IDE for a week. The results show that the RSS did actually increase for all percentiles.

See https://github.com/haskell/haskell-language-server/pull/1996 for a more localised approach.

PR #345, opened 2021-07-06
Add compatibility with GHC 9.2

fendor:
`ConP` has a new field. To avoid the usage of CPP, prefer the smart constructor `conP` which seems to be backwards compatible.

Note, that `lsp` only compiles with `head.hackage` right now, so this PR is hard to test locally.


wz1000:
Perhaps you can add `lsp` to `head.hackage`?

fendor:
Yeah I am planning on that

PR #350, opened 2021-08-02
Add dependen-sum lower version to fix build without cabal.project

### LINKED ISSUE #349: Hackage ci fails to build lsp-types-1.3.0.0
From https://github.com/haskell/lsp/issues/348#issuecomment-890574810 by @bkomuves

----------------------

Both the Hackage CI and the Matrix Builder (I'm not sure if these are separate entities or not?) fails to build lsp-types-1.3.0.0: 

* https://matrix.hackage.haskell.org/#/package/lsp-types/
* https://hackage.haskell.org/package/lsp-types-1.3.0.0/reports/

which in turn also means that there is no browsable documentation on Hackage.

Since it builds locally (with ghc 8.6.5 in my case), I would guess that the reason is maybe that some dependencies have a not-strict-enough version bound?

jneira:
* F.e. the hackage ci build
* Fixes #349

### DIFF
--- a/lsp-types/lsp-types.cabal
+++ b/lsp-types/lsp-types.cabal
@@ -89,7 +89,9 @@ library
                      , rope-utf16-splay >= 0.3.1.0
                      , scientific
                      , some
-                     , dependent-sum-template
+                     , dependent-sum-template >= 0.1.0.0
+                     -- transitive dependency of the previous one, which does not have the correct lower bound
+                     , dependent-sum >= 0.7.1.0
                      , text
                      , template-haskell
                      , temporary


jneira:
I have to test it locally, as ci here is using cabal.project

jneira:
i tried this locally, removing the source-repository-package which make the build work and the build ended succesfully

jneira:
We should try it with hls too before making a hackage revision

jneira:
hls máster cant be built with this pr nor the hackage version due to duplicate definitions for length, _start and maybe someone else
I am preparing a pr to make his work on master/hackage released version, but using this did not fail at the solver phase, so I think we could merge it

pepeiborra (approved):
Thanks for fixing this @jneira !

pepeiborra:
Could you make a Hackage revision too?

pepeiborra:
> hls master cant be built with this pr nor the hackage version due to duplicate definitions for length, _start and maybe someone else
> I am preparing a pr to make hls work on master/hackage released version, but using this did not fail at the solver phase, so I think we could merge it

Are you sure HLS master doesn't build? What error do you get?

jneira:
Oh, thanks for noting it, the work is already done with https://github.com/haskell/haskell-language-server/pull/2059 😅

PR #352, opened 2021-08-07
Sync lsp-types.cabal cabal-version with lsp.cabal

jneira:
* For make both consistent and cause i guess that could be the reason lsp-types is not built for newer ghc's in the hackage matrix: https://matrix.hackage.haskell.org/#/package/lsp-types but lsp is https://matrix.hackage.haskell.org/#/package/lsp


jneira:
It seems it is not related with the matrix versions bu i still think is a good idea to be consistent between bot packages

PR #360, opened 2021-10-30
Aeson 2 compatibility

### LINKED ISSUE #356: Aeson >= 2.0.0.0 causes a build failure
```
$ cabal install lsp-types
...
src/Language/LSP/Types/Common.hs:58:39: error:
    * Couldn't match type `Data.Aeson.KeyMap.KeyMap Value'
                     with `HashMap.HashMap k0 v0'
      Expected type: HashMap.HashMap k0 v0
        Actual type: Object
    * In the first argument of `HashMap.null', namely `o'
      In the expression: HashMap.null o
      In a stmt of a pattern guard for
                     an equation for `parseJSON':
        HashMap.null o
   |
58 |   parseJSON (Object o) | HashMap.null o = pure Empty
   |                                       ^
cabal: Failed to build lsp-types-1.3.0.1. See the build log above for details.
```
It looks like the type of Objects has changed.

michaelpj:
We get compatibility with both <2 and >=2 by using only functions that
appear in both, which don't make assumptions about the structure of
objects.

Fixes #356

### DIFF
--- a/lsp-types/src/Language/LSP/Types/Common.hs
+++ b/lsp-types/src/Language/LSP/Types/Common.hs
@@ -9,7 +9,6 @@ module Language.LSP.Types.Common where
 import Control.Applicative
 import Control.DeepSeq
 import Data.Aeson
-import qualified Data.HashMap.Strict as HashMap
 import GHC.Generics
 
 -- | A terser, isomorphic data type for 'Either', that does not get tagged when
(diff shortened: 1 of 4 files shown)


jneira (approved):
Looks great and elegant with no CPP conditions (with only some scary lens-like operators in exchange :-P) 

@fendor do you think we could remove cpp conditions from hie-bios as well?

michaelpj:
Oh and just to confirm: I built this with both versions 1.5.6 and 2.0.1, so it does definitely build with both.

fendor:
> Looks great and elegant with no CPP conditions (with only some scary lens-like operators in exchange :-P)
> 
> @fendor do you think we could remove cpp conditions from hie-bios as well?

Sounds like a good idea, will try it

PR #361, opened 2021-10-30
Fix the Semigroup instance for MarkupContent

### LINKED ISSUE #328: Semigroup instance for MarkupContent is wrong
Markdown and plaintext have different semantics wrt newlines. The current semigroup instance simply coerces plaintext to markdown, but this breaks newlines. A more correct solution would be to do something like this for any plaintext arguments that are being transformed into markdown:

```
forceMarkdownNewlines :: String -> String
forceMarkdownNewlines = unlines . fmap (<> "  ") . lines
```

michaelpj:
Fixes #328.

### DIFF
--- a/lsp-types/src/Language/LSP/Types/MarkupContent.hs
+++ b/lsp-types/src/Language/LSP/Types/MarkupContent.hs
@@ -12,6 +12,7 @@ module Language.LSP.Types.MarkupContent where
 import           Data.Aeson
 import           Data.Aeson.TH
 import           Data.Text                                      (Text)
+import qualified Data.Text as T
 import           Language.LSP.Types.Utils
 
 -- |  Describes the content type that a client supports in various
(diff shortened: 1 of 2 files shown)


lsp-types/src/Language/LSP/Types/MarkupContent.hs
```diff
@@ -81,10 +82,18 @@ sectionSeparator = "* * *\n"
 
 -- ---------------------------------------------------------------------
 
+-- | Given some plaintext, convert it into some equivalent markdown text.
+-- This is not *quite* the identity function.
+plainTextToMarkdown :: Text -> Text
+-- Line breaks in markdown paragraphs are ignored unless the line ends with two spaces.
```
pepeiborra, line 88: TIL

pepeiborra (approved):

PR #362, opened 2021-10-30
Add `window/showDocument` and some missing capabilities

michaelpj:
Noticed while looking into https://github.com/haskell/lsp/issues/329 (includes that one).


jneira:
So maybe could we check some of the missing lsp spec features in #134

PR #363, opened 2021-11-02
Add mergify config

michaelpj:
Copied from HLS.


jneira (approved):
thanks for bringing it here

PR #366, opened 2021-11-06
Use appropriate number types

### LINKED ISSUE #354: Provide newtype wrappers for new number types of lsp spec 3.16
From https://github.com/haskell/lsp/issues/134#issuecomment-914558600 by @pranaysashank  (thanks!)

> Since 3.16, there are new number types https://microsoft.github.io/language-server-protocol/specifications/specification-current/#number . We may want to provide newtype wrappers for them as using Int may lead to subtle bugs like haskell/haskell-language-server#2169

michaelpj:
This ensures that we use either:
- `Int32`
- `Word32`
- `Float`

In particular, this gets us:
- Appropriate `Bounded` instances (see the original issue
haskell/haskell-language-server#2169).
- More picky `aeson` instances for the bounded types.

Rather than use newtypes, we just use the existing appropriate Haskell
numeric types for bounded integers.

Fixes #354.

NOTE: I also bumped the `lsp-types` major version, since this is a type-representation change.


jneira:
Many thanks for fixing this. Would need it some work to adapt clients (like hls)? could we do something here to make easier the upgrade?

michaelpj:
Yeah, it'll require some changes, but just a scattering of `fromIntegral`. I don't know of a way to avoid that, unfortunately: ultimately we *are* changing the representation of the types.

lsp-types/src/Language/LSP/Types/DocumentColor.hs
```diff
@@ -45,10 +45,10 @@ deriveJSON lspOptions ''DocumentColorParams
 -- | Represents a color in RGBA space.
 data Color =
   Color
-    { _red   :: Int -- ^ The red component of this color in the range [0-1].
-    , _green :: Int -- ^ The green component of this color in the range [0-1].
-    , _blue  :: Int -- ^ The blue component of this color in the range [0-1].
-    , _alpha :: Int -- ^ The alpha component of this color in the range [0-1].
+    { _red   :: Float -- ^ The red component of this color in the range [0-1].
```
jneira, line 48: Surprised to see a Float is needed for that, but i guess it is from the spec

lsp-types/src/Language/LSP/Types/Utils.hs
```diff
@@ -1,6 +1,7 @@
 {-# LANGUAGE RankNTypes                 #-}
 {-# LANGUAGE TemplateHaskell            #-}
 {-# LANGUAGE LambdaCase                 #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
```
jneira, line 4: Is this needed? There is no other changes in the module

lsp-types/src/Language/LSP/Types/WorkspaceEdit.hs
```diff
@@ -359,6 +359,11 @@ data ApplyWorkspaceEditResponseBody =
       -- logging or to provide a suitable error for a request that
       -- triggered the edit.
     , _failureReason :: Maybe Text
+      -- | Depending on the client's failure handling strategy `failedChange`
+      -- might contain the index of the change that failed. This property is
+      -- only available if the client signals a `failureHandling` strategy
+      -- in its client capabilities.
+    , _failedChange :: Maybe Word32
```
jneira, line 366: This is a not related change, maybe it worths to note it as a separate commit or in another way

lsp-types/src/Language/LSP/VFS.hs
```diff
@@ -278,15 +279,6 @@ closeVFS vfs (J.NotificationMessage _ _ params) =
...
 -- ---------------------------------------------------------------------
-{-
-
-data TextDocumentContentChangeEvent =
-  TextDocumentContentChangeEvent
-    { _range       :: Maybe Range
-    , _rangeLength :: Maybe Int
-    , _text        :: String
-    } deriving (Read,Show,Eq)
--}
```
jneira, line 289: Why is it is removed? It is included elsewhere or was being unused?

lsp/test/JsonSpec.hs
```diff
@@ -7,6 +7,9 @@
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 -- For the use of MarkedString
 {-# OPTIONS_GHC -fno-warn-deprecations #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
```
jneira, line 12: More pragmas with no module changes, maybe they are needed for upstream changes?

lsp-types/src/Language/LSP/Types/DocumentColor.hs
michaelpj: I think that's the best thing to correspond to a "decimal"? WDYT?

lsp-types/src/Language/LSP/Types/WorkspaceEdit.hs
michaelpj: Oops thanks, just a small thing I noticed we missed!

lsp-types/src/Language/LSP/VFS.hs
michaelpj: It's commented out

lsp-types/src/Language/LSP/Types/DocumentColor.hs
jneira: nvm, i did not understand the usage

lsp-types/src/Language/LSP/VFS.hs
jneira: oh yeah 🤦

jneira (approved):
LGTM, many thanks for fixing that

PR #377, opened 2021-12-15
WIP: Try to extend CI to 9.2.1

michaelpj:


michaelpj:
Looks like we're blocked on at least:
- https://github.com/obsidiansystems/dependent-sum/issues/66
- https://github.com/obsidiansystems/constraints-extras/issues/45

michaelpj:
I've pulled out the bounds changes to another PR, but this won't work for a bit.

Ailrun:
@michaelpj For `constraint-extras`, it looks like their HEAD already supports GHC 9.2, but the actual release haven't been cut yet. (At least their CI works: https://github.com/obsidiansystems/constraints-extras/runs/4389445166?check_suite_focus=true)

Bodigrim:
Looking at this and similar tickets for HLS, it seems that `dependent-map` subtree (`dependent-sum`, `constraints-extras`, `dependent-sum-template`, `th-extras`) causes constant delays, is it? Could it be replaced with `typerep-map`? Or maybe roll out a hand-written implementation of a type, isomorphic to `DMap SMethod`?

michaelpj:
Yes, I think that's a good suggestion. I'd wondered about doing that, but I don't currently have a handle on how it's used. I'll take a look at what you suggested.

michaelpj:
Removing `dependent-sum` and friends isn't trivial. We use `GEq` in a few places apart from the dependent map in `lsp`, and while we could just depend on `some` to get `GEq`, I really wouldn't want to write the instances by hand (big datatypes!), which pushes us back into the arms of `dependent-sum-template`.

michaelpj:
Okay, now I think it's worth having a look at whether we can indeed get rid of `dependent-map`. It seems that we use it "properly" in a way that can't be obviously replaced by `typerep-map`. So hand-rolling a simple alternative might be the way to go.

jneira:
> Okay, now I think it's worth having a look at whether we can indeed get rid of `dependent-map`. It seems that we use it "properly" in a way that can't be obviously replaced by `typerep-map`. So hand-rolling a simple alternative might be the way to go.

we could leverage changes in lsp? or hls ones are highly unrelated with?

michaelpj:
I don't think I could get *rid* of the usage in `lsp`, or at least I don't understand it enough to say yet. So easier to replace the datastructure at the moment.

michaelpj:
Done in https://github.com/haskell/lsp/pull/384

PR #378, opened 2021-12-15
Bounds and project tidying

michaelpj:
Pulled out of https://github.com/haskell/lsp/pull/377

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -20,7 +20,7 @@ jobs:
     - uses: haskell/actions/setup@v1
       with:
         ghc-version: ${{ matrix.ghc }}
-        cabal-version: '3.4'
+        cabal-version: '3.6'
 
     - name: Cache
       uses: actions/cache@v1
--- a/lsp-test/lsp-test.cabal
+++ b/lsp-test/lsp-test.cabal
@@ -99,7 +99,7 @@ test-suite func-test
   main-is:             FuncTest.hs
   hs-source-dirs:      func-test
   type:                exitcode-stdio-1.0
-  build-depends:       base <4.16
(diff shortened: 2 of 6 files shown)


lsp-test/lsp-test.cabal
```diff
@@ -99,7 +99,7 @@ test-suite func-test
   main-is:             FuncTest.hs
   hs-source-dirs:      func-test
   type:                exitcode-stdio-1.0
-  build-depends:       base <4.16
+  build-depends:       base
```
jneira, line 102: should not have the package base bounds?

.github/workflows/haskell.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     - uses: haskell/actions/setup@v1
       with:
         ghc-version: ${{ matrix.ghc }}
-        cabal-version: '3.4'
+        cabal-version: 'latest'
```
jneira, line 23: not always last cabal is good, would `3.6` be too much careful?

jneira (approved):
look good overall, thanks for the cleaning up!

.github/workflows/haskell.yml
jneira: afaik 3.6 will pick 3.6.2.0, 3.6.0.0 suffered some bugs

michaelpj: Fair enough.

lsp-test/lsp-test.cabal
michaelpj: It depends on `lsp-test` which has the bounds - adding them here just means extra work IMO.

jneira: oh I miss the existing bound in the package lib

PR #379, opened 2021-12-15
Simplify down to one stack.yaml

michaelpj:
As discussed on Matrix.

Since we don't use this for user installation, nor for testing multiple
versions of GHC (we have Haskell-CI based on `cabal` for that), the only
use of this is to provide a dev setup for stack users. As such, one
`stack.yaml` should be sufficient.

I checked and `stack build` still succeeds.


michaelpj:
Bumping the install-nix-action version seems to fix the nix issue, the bump is in https://github.com/haskell/lsp/pull/378, I'll rebase this after taht's in.

jneira (approved):
nice! the remaining stack.yaml is almost perfect

stack.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+packages:
+- ./lsp
+- ./lsp-types
+- ./lsp-test
+
+flags: {}
+extra-package-dbs: []
+nix:
+  packages: [icu]
```
Ailrun, line 11: Wow, how clean it is! Thanks!

Ailrun (approved):

michaelpj:
It's only clean because 8.10 is so old now that lots of packages have got into stackage :P

jneira:
...and LSP packages and it's deps live in stackage

PR #380, opened 2021-12-15
Update GHC build matrix

michaelpj:
Update to be in line with HLS, except with only one 8.10 version.
Maybe we could get rid of even more, but good to match HLS.


jneira (approved):
lgtm, thanks again

michaelpj:
@jneira can you change the branch protection settings since the set of required checks is different now :pray:

jneira:
> @jneira can you change the branch protection settings since the set of required checks is different now 🙏

you should have the rights to do it now 😄

michaelpj:
Done.

PR #382, opened 2021-12-20
Replace Word32 with a newtype

michaelpj:
This time with the correct (strange) bounds.

### DIFF
--- a/lsp-types/src/Language/LSP/Types/Common.hs
+++ b/lsp-types/src/Language/LSP/Types/Common.hs
@@ -1,6 +1,9 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE DataKinds                  #-}
+{-# LANGUAGE DeriveAnyClass             #-}
 {-# LANGUAGE DeriveTraversable          #-}
 {-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE DerivingStrategies         #-}
 {-# LANGUAGE TypeOperators              #-}
 
 -- | Common types that aren't in the specification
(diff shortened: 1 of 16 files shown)


michaelpj:
Builds are generally passing in https://github.com/haskell/haskell-language-server/pull/2494, so I think this is safe to merge, and then we should be okay to do a release and then fix up HLS.

Bodigrim:
Introducing `OddWord` dependency is likely to cause a trouble: it was not updated in years and already does not build with GHC 9.2.

michaelpj:
> Introducing OddWord dependency is likely to cause a trouble: it was not updated in years and OddWord-1.0.2.0 already does not build with GHC 9.2.

Ugh, how annoying. I was hoping that it was a simple enough package that it would just "keep working".

I was really hoping to avoid writing the plumbing for a 31-bit word myself...

Bodigrim:
Unexpectedly, older releases of `OddWord` does work (I guess they contain less type level magic), and since I've revised `OddWord-1.0.2.0` with an appropriate `base` bound,  `cabal` should now be able to derive a working build plan for GHC 9.2. Dunno, still does not sound sustainable to me, but I've been just passing by.

michaelpj:
Maybe I'll ask if I can take over `OddWord`. It's a small package.

jneira (approved):
looks good to me, do you think we can continue using OddWord?
Well if it becomes a blocker we can always do the boilerplate

michaelpj:
Yes, ripping it out later would be a localized change. Maybe the LSP people will hear my plea and change the upper bound (yeah right).

Bodigrim:
Any particular reason to prefer `OddWord` to any well-maintained package for modular arithmetic? Something like https://hackage.haskell.org/package/mod-0.1.2.2/docs/Data-Mod-Word.html would do, for instance. `OddWord` is not even on Stackage.

michaelpj:
Thanks, it didn't occur to me to look for a modular arithmetic package, that makes a lot of sense in fact!

`Mod` is missing a bunch of the numeric instances, but that might be okay. Let me see.

lsp-types/src/Language/LSP/Types/Common.hs
```diff
@@ -10,14 +13,41 @@ module Language.LSP.Types.Common (
...
-import Data.Word (Word32)
-import GHC.Generics
+import Data.Mod.Word
+import Text.Read (Read(readPrec))
+import GHC.Generics hiding (UInt)
+
+-- | The "uinteger" type in the LSP spec.
+--
+-- Unusually, this is a **31**-bit unsigned integer, not a 32-bit one.
+newtype UInt = UInt (Mod 31)
```
Bodigrim, line 31:
```suggestion
newtype UInt = UInt (Mod 2147483648)
```
Or `2 ^ 31`, if you import `GHC.TypeNats`.

lsp-types/src/Language/LSP/Types/Common.hs
```diff
@@ -10,14 +13,41 @@ module Language.LSP.Types.Common (
...
+  deriving anyclass (NFData)
+
+instance Read UInt where
+  readPrec = fromInteger <$> readPrec
+
+instance Real UInt where
+  toRational (UInt u) = toRational $ unMod u
+
+instance Integral UInt where
+  quotRem (UInt x) (UInt y) = coerce $ E.quotRem x y
```
Bodigrim, line 43:
```suggestion
  quotRem (UInt x) (UInt y) = coerce $ fromIntegral $ quotRem (unMod x) (unMod y)
```
`E.quotRem` (division in a modular ring) is not quite what you want here.

michaelpj: Hmm, I mistakenly thought it would be equivalent, thank you.

lsp-types/src/Language/LSP/Types/Common.hs
michaelpj: :facepalm: 

Fortunately, the HLS tests do fail with this, so I was going to notice eventually...

PR #383, opened 2021-12-21
Eliminate dependent-sum-template dependency

Bodigrim:
Cf. https://github.com/haskell/lsp/pull/377#issuecomment-998869748

Eliminating `dependent-sum-template`, `th-extras` and `th-abstraction` should make builds more robust and provide a better experience when upgrading to new GHCs.

### DIFF
--- a/lsp-types/lsp-types.cabal
+++ b/lsp-types/lsp-types.cabal
@@ -90,8 +90,6 @@ library
                      , rope-utf16-splay >= 0.3.1.0
                      , scientific
                      , some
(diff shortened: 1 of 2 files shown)


jneira (approved):
wow, many many thanks, I had to chase maintainers of those packages several times to get support for newer ghcs
will wait for @michaelpj thoughts to merge

michaelpj:
Nice! Using unsafeCoerce there wouldn't have occurred to me, but it seems entirely fine, and avoids the instance becoming huge. LGTM.

michaelpj (approved):

PR #384, opened 2021-12-22
Eliminate dependent-sum and dependent-map dependencies

Bodigrim:
As discussed in https://github.com/haskell/lsp/pull/377#issuecomment-999452278

This removes `dependent-map`, `dependent-sum`, `constraints-extra` and `constraints` from dependencies and makes `lsp` buildable with GHC 9.2.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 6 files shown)


jneira (approved):
it does not add too much code and it looks good, so it totally worths the drop of deps, many thanks!

michaelpj:
Amazing, thanks for doing this!

PR #388, opened 2022-01-03
fix textDocument/selectionRange's method

### LINKED ISSUE #387: textDocument/selectionRange uses STextDocumentFoldingRange
While trying to implement haskell/haskell-language-server#212, I notice that in `lsp-types`, `textDocument/selectionRange` points to `STextDocumentFoldingRange`.
https://github.com/haskell/lsp/blob/08e8a72c4f1238d6a49f7e199962cc89969a5c2a/lsp-types/src/Language/LSP/Types/Method.hs#L322-L323
Is this a typo?

kokobd:
fix #387

### DIFF
--- a/lsp-types/src/Language/LSP/Types/Method.hs
+++ b/lsp-types/src/Language/LSP/Types/Method.hs
@@ -320,7 +320,7 @@ instance FromJSON SomeClientMethod where
   parseJSON (A.String "textDocument/rename")                 = pure $ SomeClientMethod STextDocumentRename
   parseJSON (A.String "textDocument/prepareRename")          = pure $ SomeClientMethod STextDocumentPrepareRename
   parseJSON (A.String "textDocument/foldingRange")           = pure $ SomeClientMethod STextDocumentFoldingRange
-  parseJSON (A.String "textDocument/selectionRange")         = pure $ SomeClientMethod STextDocumentFoldingRange
+  parseJSON (A.String "textDocument/selectionRange")         = pure $ SomeClientMethod STextDocumentSelectionRange
   parseJSON (A.String "textDocument/prepareCallHierarchy")   = pure $ SomeClientMethod STextDocumentPrepareCallHierarchy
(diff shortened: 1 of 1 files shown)


jneira (approved):
Good catch, it should be catched by some type of testing, but it is a generic issue we should tackle somewhere else

kokobd:
Nix build passes but direct build fails. It looks like an orphan instance in `JsonSpec.hs` conflicts with latest version of aeson.
https://github.com/haskell/lsp/blob/08e8a72c4f1238d6a49f7e199962cc89969a5c2a/lsp/test/JsonSpec.hs#L162
Should it be fixed here or in another PR?

kokobd:
According to aeson's [changlog](https://github.com/haskell/aeson/blob/master/changelog.md), it provides `Arbitrary` instance for `Value` starting from version `2.0.3.0`

jneira:
wow, @michaelpj got to support aeson >= 2.0 without any cpp, not sure if we could avoid it fixing this

jneira:
Add a instance is a breaking change and it shouldbe a major version bump 🙄 
Well i would update the aseon bounds to exclude 2.0.3.0 to make ci work and check the pr changes and left the fix for another pr (if @michaelpj agree)

kokobd:
@jneira I've added version bound by another commit in this PR. CI has passed now.

michaelpj:
Ugh, thanks for fixing this.

Re: the CPP, I think there's a more elegant solution. We only need the instance in the tests for the `lsp` package. So we can put a `aeson >= 2.0.3.0` constraint on the _test_ component of `lsp`, and unconditionally delete the instance. That will force a new `aeson`... but only for people who build the test component. Users who only depend on the library won't get that constraint and will continue being able to use any vesrion.

michaelpj:
(Incidentally, I don't love the fact that `merge-me` auto-merges even if someone pushes another commit!)

michaelpj:
https://github.com/haskell/lsp/pull/389

Bodigrim:
> Add a instance is a breaking change and it should be a major version bump

This is a non-orphan instance, which according to PVP is non-breaking.

PR #389, opened 2022-01-04
Remove CPP by using higher bound for the lsp test component

michaelpj:
This gets us the instance that we need. Users who just depend on the
library component won't be affected by this and can continue to build
with any aeson version.


jneira (approved):
lgtm, neat trick

PR #392, opened 2022-01-14
Support text-2.0

### LINKED ISSUE #391: Migration to text 2.0
`lsp` uses `rope-utf16-splay` to represent buffers:

https://github.com/haskell/lsp/blob/1468d13d6ef3ef18b34002801012ba2139bbd9da/lsp-types/lsp-types.cabal#L91

This is not a good fit since `text-2.*` uses an UTF-8 encoding. We can either create an equivalent data structure for UTF-8, or drop it and work with `Text` directly. @alanz do you know how critical the rope data structure is for performance?

Bodigrim:
Closes #391. 

This is a draft, need to handle `fromJust` more gracefully (and publish `text-lines` on Hackage).

I tried adapting `rope-utf16-splay` to `text-2.0`, but did not succeed: the changes required are too radical to its API. Hence a switch to `text-lines`.

### DIFF
--- a/lsp-types/src/Language/LSP/VFS.hs
+++ b/lsp-types/src/Language/LSP/VFS.hs
@@ -52,8 +52,8 @@ import           Data.Ord
 import qualified Data.HashMap.Strict as HashMap
 import qualified Data.Map.Strict as Map
 import           Data.Maybe
-import           Data.Rope.UTF16 ( Rope )
-import qualified Data.Rope.UTF16 as Rope
+import           Data.Text.Utf16.Rope ( Rope )
+import qualified Data.Text.Utf16.Rope as Rope
 import qualified Language.LSP.Types           as J
 import qualified Language.LSP.Types.Lens      as J
 import           System.FilePath
@@ -136,7 +136,7 @@ applyCreateFile (J.CreateFile uri options _ann) =
   updateVFS $ Map.insertWith 
                 (\ new old -> if shouldOverwrite then new else old)
                 (J.toNormalizedUri uri)
-                (VirtualFile 0 0 (Rope.fromText ""))
+                (VirtualFile 0 0 mempty)
   where 
     shouldOverwrite :: Bool 
     shouldOverwrite = case options of 
@@ -260,7 +260,7 @@ persistFileVFS vfs uri =
           action = do
             exists <- doesFileExist tfn
             unless exists $ do
-               let contents = Rope.toString (_text vf)
+               let contents = T.unpack (Rope.toText (_text vf))
                    writeRaw h = do
                     -- We honour original file line endings
                     hSetNewlineMode h noNewlineTranslation
@@ -291,26 +291,18 @@ applyChanges = foldl' applyChange
 applyChange :: Rope -> J.TextDocumentContentChangeEvent -> Rope
 applyChange _ (J.TextDocumentContentChangeEvent Nothing Nothing str)
   = Rope.fromText str
-applyChange str (J.TextDocumentContentChangeEvent (Just (J.Range (J.Position sl sc) _to)) (Just len) txt)
-  = changeChars str start (fromIntegral len) txt
-  where
-    start = Rope.rowColumnCodeUnits (Rope.RowColumn (fromIntegral sl) (fromIntegral sc)) str
-applyChange str (J.TextDocumentContentChangeEvent (Just (J.Range (J.Position sl sc) (J.Position el ec))) Nothing txt)
-  = changeChars str start len txt
-  where
-    start = Rope.rowColumnCodeUnits (Rope.RowColumn (fromIntegral sl) (fromIntegral sc)) str
-    end = Rope.rowColumnCodeUnits (Rope.RowColumn (fromIntegral el) (fromIntegral ec)) str
-    len = end - start
+applyChange str (J.TextDocumentContentChangeEvent (Just (J.Range (J.Position sl sc) (J.Position fl fc))) _ txt)
+  = changeChars str (Rope.Position (fromIntegral sl) (fromIntegral sc)) (Rope.Position (fromIntegral fl) (fromIntegral fc)) txt
 applyChange str (J.TextDocumentContentChangeEvent Nothing (Just _) _txt)
   = str
 
 -- ---------------------------------------------------------------------
 
-changeChars :: Rope -> Int -> Int -> Text -> Rope
-changeChars str start len new = mconcat [before, Rope.fromText new, after']
+changeChars :: Rope -> Rope.Position -> Rope.Position -> Text -> Rope
+changeChars str start finish new = mconcat [before', Rope.fromText new, after]
   where
-    (before, after) = Rope.splitAt start str
-    after' = Rope.drop len after
+    (before, after) = fromJust $ Rope.splitAtPosition finish str
+    (before', _) = fromJust $ Rope.splitAtPosition start before
 
 -- ---------------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


jneira:
Thanks for the pr. For things like this i miss some sort of benchmark here. We could use the hls benchmark making ghcide use this pr though.

michaelpj:
Nice!

Will `text-lines` be compatible with text both pre- and post-2? That would be ideal, so we don't force a `text-2` dependency if we can help it.

Bodigrim:
> For things like this i miss some sort of benchmark here. We could use the hls benchmark making ghcide use this pr though.

Could you point me to a specific benchmark please?

> Will `text-lines` be compatible with text both pre- and post-2?

Of course, it is compatible with both already.

jneira:
> > For things like this i miss some sort of benchmark here. We could use the hls benchmark making ghcide use this pr though.
> 
> Could you point me to a specific benchmark please?

Sure, it is the benchmark we use for ghcide: https://github.com/haskell/haskell-language-server/blob/9c2bc32875ae0bb6871d7cc06547da129cf4ff5f/ghcide/ghcide.cabal#L449 in [our ci](https://github.com/haskell/haskell-language-server/blob/master/.github/workflows/bench.yml)

lsp-types/src/Language/LSP/VFS.hs
```diff
@@ -291,26 +291,18 @@ applyChanges = foldl' applyChange
...
 
-changeChars :: Rope -> Int -> Int -> Text -> Rope
-changeChars str start len new = mconcat [before, Rope.fromText new, after']
+changeChars :: Rope -> Rope.Position -> Rope.Position -> Text -> Rope
+changeChars str start finish new = mconcat [before', Rope.fromText new, after]
   where
-    (before, after) = Rope.splitAt start str
-    after' = Rope.drop len after
+    (before, after) = fromJust $ Rope.splitAtPosition finish str
+    (before', _) = fromJust $ Rope.splitAtPosition start before
```
Bodigrim, line 305: What's the best way to handle malformed input here? Since LSP counts by UTF-16 code units (not code points), it is possible that split happens in the middle of a code point, in which case `text-lines` returns `Nothing`. 
(`rope-utf16-splay` in such situation silently extends split region until the end of a code point)

lsp-types/src/Language/LSP/VFS.hs
```diff
@@ -341,9 +333,8 @@ getCompletionPrefix pos@(J.Position l c) (VirtualFile _ _ ropetext) =
             lastMaybe [] = Nothing
             lastMaybe xs = Just $ last xs
 
-        curLine <- headMaybe $ T.lines $ Rope.toText
-                             $ fst $ Rope.splitAtLine 1 $ snd $ Rope.splitAtLine (fromIntegral l) ropetext
-        let beforePos = T.take (fromIntegral c) curLine
+        let curRope = fst $ Rope.splitAtLine 1 $ snd $ Rope.splitAtLine (fromIntegral l) ropetext
+        beforePos <- Rope.toText . fst <$> Rope.splitAt (fromIntegral c) curRope
```
Bodigrim, line 337: BTW this fixes an existing bug: `T.take` counts code points, while LSP requires counting UTF-16 code units.

Bodigrim:
Benchmarks adapted from https://github.com/ollef/rope-bench/blob/main/bench/Bench.hs show that `text-lines` is at least twice faster than `rope-utf16-splay` for `splitAt`. Plus `splitAtLine` takes logarithmic time instead of linear.

```
All
  inputlen 1000, 1000 iterations
    linear mods
      mods sized 10
        SplayTree: OK (1.35s)
          1.30 ms ± 120 μs
        TextLines: OK (1.29s)
          627  μs ±  54 μs
    random mods
      mods sized 10
        SplayTree: OK (3.03s)
          1.48 ms ±  97 μs
        TextLines: OK (2.60s)
          644  μs ±  52 μs
    toText . fromText
      SplayTree:   OK (3.56s)
        846  ns ±  37 ns
      TextLines:   OK (4.43s)
        133  ns ± 5.0 ns
  inputlen 10000, 1000 iterations
    linear mods
      mods sized 10
        SplayTree: OK (18.24s)
          2.12 ms ± 211 μs
        TextLines: OK (2.35s)
          575  μs ±  30 μs
    random mods
      mods sized 10
        SplayTree: OK (2.37s)
          2.28 ms ± 218 μs
        TextLines: OK (1.22s)
          1.19 ms ± 114 μs
    toText . fromText
      SplayTree:   OK (2.64s)
        10.1 μs ± 512 ns
      TextLines:   OK (1.58s)
        374  ns ±  31 ns
  inputlen 100000, 1000 iterations
    linear mods
      mods sized 10
        SplayTree: OK (2.31s)
          2.24 ms ± 119 μs
        TextLines: OK (5.38s)
          659  μs ±  60 μs
    random mods
      mods sized 10
        SplayTree: OK (2.45s)
          4.80 ms ± 383 μs
        TextLines: OK (1.89s)
          1.84 ms ± 114 μs
    toText . fromText
      SplayTree:   OK (3.56s)
        106  μs ± 5.5 μs
      TextLines:   OK (1.65s)
        3.12 μs ± 248 ns
```

lsp-types/src/Language/LSP/VFS.hs
michaelpj: @alanz ? do you remember what's up here?

michaelpj:
@Bodigrim do you feel motivated to maintain `text-lines` indefinitely? Performance is great, but I am also wary of adding another thing to your list. Since the existing library got fixed we could just use that.

lsp-types/src/Language/LSP/VFS.hs
alanz: Nope. To be honest I have never really paid attention to issues of splits in the middle of code points, assuming the client will send meaningful input, since they are managing a cursor over the text and sending its positions.

michaelpj: We *could* throw. Maybe blowing up with an informative error message is the most sensible thing to do here.

Bodigrim:
I think I'm beyond the point of no return :) and will release and maintain it anyway.

More benchmarks, now with a realistic payload mimicking `lsp`:
```
text-1.2:
  TextLines: 
    2.15 ms ±  53 μs
  RopeSplay: 
    6.06 ms ± 181 μs

text-2.0:
  TextLines: 
    3.71 ms ± 125 μs
  RopeSplay: 
    11.3 ms ± 371 μs
```

ollef:
Hard to argue with those performance numbers. :+1:

lsp-types/src/Language/LSP/VFS.hs
```diff
@@ -357,6 +348,7 @@ getCompletionPrefix pos@(J.Position l c) (VirtualFile _ _ ropetext) =
             let modParts = dropWhile (not . isUpper . T.head)
                                 $ reverse $ filter (not .T.null) xs
                 modName = T.intercalate "." modParts
+            curLine <- headMaybe $ T.lines $ Rope.toText curRope
```
michaelpj, line 351: Why not use `Rope.lines`? Won't that be faster?

lsp/test/VspSpec.hs
```diff
@@ -65,7 +64,7 @@ vspSpec = do
           ]
         new = applyChange (fromString orig)
                 $ J.TextDocumentContentChangeEvent (Just $ J.mkRange 2 1 2 5) (Just 4) ""
-      lines (Rope.toString new) `shouldBe`
+      lines (T.unpack $ Rope.toText new) `shouldBe`
```
michaelpj, line 67: I think these could all be 
```suggestion
      Rope.lines new `shouldBe`
```

lsp-types/src/Language/LSP/VFS.hs
michaelpj: I would be happy with either:
- Throwing, based on judging this to be very unlikely.
- Returning `Nothing`, and propagating that up to the exported functions. In `changeFromClientVFS` we can drop the change and log, as is done in the existing `Nothing` branch.

Thoughts @alanz ?

michaelpj (approved):
LGTM, so long as we resolve the question of what to do with malformed input.

lsp-types/src/Language/LSP/VFS.hs
michaelpj: Alternatively, we can merge this with the `fromJust` and experiment with removing it with @Bodigrim out of the loop :)

michaelpj: Looking more at the code, I think we could push these failures right the way up to the LSP message response, which is probably the right thing to do, but I think we should do it separately. So my vote goes for "leave `fromJust` for now and remove subsequently", which I volunteer to look into.

lsp/test/VspSpec.hs
Bodigrim: Excellent idea!

lsp-types/src/Language/LSP/VFS.hs
Bodigrim: `curRope` is already a single line, but it likely includes an enclosing `\n`. The only purpose of `headMaybe . T.lines . Rope.toText` is to strip this `\n`. I've changed it to a more explicit version.

lsp-types/src/Language/LSP/VFS.hs
alanz: I think this is unlikely to happen in practice, so leaving as is would likely be safe, as is pushing it up to an LSP response. For the latter it at least makes us aware it is happening, and hopefully that will show up in initial developer dogfooding, or not at all.

PR #396, opened 2022-01-30
fix stack build by adding text-rope-0.1 into stack extra-deps

### LINKED ISSUE #395: stack build error
> Error: While constructing the build plan, the following exceptions were encountered:
> 
> In the dependencies for lsp-types-1.4.0.1:
>     text-rope needed, but the stack configuration has no specified version  (latest matching version is 0.1)
> needed since lsp-types is a build target.
> 
> Some different approaches to resolving this:
> 
>   * Recommended action: try adding the following to your extra-deps in /home/dev/files/repos/haskell-lsp/stack.yaml:
> 
> - text-rope-0.1@sha256:0654bbff798a5b32d63de75bdfce3ea9a7b9abf1e3325c620878af9d48061e33,1953
> 
> Plan construction failed.

daquexian:
Fix #395

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -9,3 +9,5 @@ flags: {}
 extra-package-dbs: []
 nix:
   packages: [icu]
+extra-deps:
+- text-rope-0.1


jneira (approved):
lgtm, as we dont have any ci job for stack we cant check it so you local test would be good, many thanks!

PR #397, opened 2022-01-30
In simple example, use LambdaCase following the suggestion of hlint

daquexian:


jneira (approved):
lgtm, the use of lambda case is pretty standars nowadays, thanks!

PR #398, opened 2022-01-30
Add build stack step

jneira:
* To check the stack.yaml does not go out of sync (see #395)

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -33,7 +33,11 @@ jobs:
 
     - name: Cabal update
       run: cabal update
-    - name: Build
+    - name: Build using cabal
       run: cabal build all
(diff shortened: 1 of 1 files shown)


jneira:
This needs a dedicated cache to make it faster

jneira:
let's see if the prev pr fixed the build

michaelpj:
Are any of our contributors using stack? We could just drop it, since it's only really useful for dev I guess.

michaelpj (approved):

jneira:
> Are any of our contributors using stack? We could just drop it, since it's only really useful for dev I guess.

well at least @daquexian tried it 🙂

and I think it could be seen the other way  around: if we keep a working stack.yaml we will have more chances of having a contributor using stack

PR #406, opened 2022-02-17
Add STM function to snapshot VFS

wz1000:
This is useful if you want to take a snapshot of the entire VFS state at a particular point.


michaelpj (approved):
Sure, why not.

PR #407, opened 2022-02-19
Add functions to convert between code-point and code-unit positions

michaelpj:
LSP `Position`s use UTF-16 code units for offsets within lines; most
other sane tools (like GHC) use Unicode code points. We need to use the
right one in the right place, otherwise we get issues like
https://github.com/haskell/haskell-language-server/issues/2646.

This is pretty unpleasant, since code points are variable-size, so you
can't do the conversion without having the file text itself.

This PR provides a type for positions using code points (for clients to
use to help them be less confused) and functions for using the VFS to
convert between those and LSP positions.


michaelpj:
@Bodigrim I attempted to follow the recipe you gave me [here](https://github.com/haskell/haskell-language-server/issues/2646#issuecomment-1026206511), but I'd appreciate a look to make sure I didn't do it wrong!

The process does seem to be quite expensive as it stands: both cases have to do a `Rope1.fromText . Rope2.toText` conversion, and going from code points to code units requires doing it twice! Awful.

michaelpj:
Ah! The UTF-16 `TextLines` is just a newtype wrapper around the other one! So these conversions should be free, but the constructor isn't exported. Either exporting it or providing some conversion functions would be very helpful!

wz1000:
Would it be useful to track if the file consists entirely of `latin-1` or whatever subset of codepoints there are such that the two notions of position are equivalent, and if so, avoid doing all the conversion work in those cases?

Bodigrim:
> Ah! The UTF-16 `TextLines` is just a newtype wrapper around the other one! So these conversions should be free, but the constructor isn't exported. Either exporting it or providing some conversion functions would be very helpful!

Please try using an unreleased version of `text-rope`. It makes `TextLines` interchangeable. 
```cabal
source-repository-package
  type: git
  location: https://github.com/Bodigrim/text-rope
```

The next step on my side would be to make a `Rope` type, allowing native indexing in both ways without conversions. The question is: do we foresee a use case, where a client is genuinely interested in UTF-16 positions? 

@wz1000 let's make it correct first, and care about optimizations later. Unless you deal with extremely long lines (not files), this should not be a bottleneck.

michaelpj:
> Would it be useful to track if the file consists entirely of latin-1

Maybe! That would also have downsides - it would be some amount of ongoing work every time the file changes.

I agree with Bodigrim, basically: let's do it this way and optimise if it turns out to be a bottleneck.

> The next step on my side would be to make a Rope type, allowing native indexing in both ways without conversions.

Well, since the rope types are interchangeable now, the current approach doesn't seem too bad. But maybe you have an idea for how to make it nicer or faster!

> The question is: do we foresee a use case, where a client is genuinely interested in UTF-16 positions?

I'm not exactly sure what you mean, but in our case this is really inescapable. The LSP spec insists that positions use UTF-16 code units, and GHC gives us positions using code points. We have to convert when the client sends us positions, and back when we send them to the client.

Maybe they'll change the spec (basically everyone hates the status quo), but I wouldn't hold my breath.

michaelpj:
> Please try using an unreleased version of text-rope. It makes TextLines interchangeable.

That looks perfect. I'm also happy to have conversion functions - perhaps you want to retain the option to change the representation so that the two modules differ.

michaelpj:
No, I had entirely confused myself, sorry @Bodigrim . The current version is using the `Rope` types, not the `TextLines` types. I need to check whether going to `TextLines` and doing the work is faster than staying with ropes but converting between them.

michaelpj:
Okay, I pushed a version using the unreleased interchangeable `TextLines`. I think this is better: we still pay a cost linear in the document to convert to `TextLines`, but we only do it once...

michaelpj:
Perhaps the thing to do would be to extract just the line in question, and then only operate on that. That would probably allow us to do something relatively stupid and not worry too much, since we're relatively unlikely to have extremely long lines.

lsp/src/Language/LSP/VFS.hs
```diff
@@ -346,6 +353,45 @@ changeChars logger str start finish new = do
...
+-- We need the file itself because this requires translating between code points and code units.
+codePointPositionToPosition :: VirtualFile -> CodePointPosition -> J.Position
+codePointPositionToPosition vFile (CodePointPosition cpl cpc) =
+  let text = _file_text vFile
+      lines = Rope.toTextLines text
+      -- Split at the given position in *code points*
+      (prefix, _) = ULines.splitAtPosition (ULines.Position (fromIntegral cpl) (fromIntegral cpc)) lines
+      -- Get the length of the prefix in *code units*
+      (Lines.Position cul cuc) = Lines.lengthAsPosition prefix
+  in J.Position (fromIntegral cul) (fromIntegral cuc)
```
Bodigrim, line 377: There are two tricks which can make it faster:
1. It's faster to split `Rope`, and convert `toTextLines` later. 
2. Line position is always preserved.

```suggestion
  let rope = _file_text vFile
      (_, rope') = Rope.splitAtLine (fromIntegral cpl)
      lines = Rope.toTextLines rope'
      -- Split at the given position in *code points*
      (prefix, _) = ULines.splitAt (fromIntegral cpc) lines
      -- Get the length of the prefix in *code units*
      cuc = Lines.length prefix
  in J.Position cpl (fromIntegral cuc)
```

This is still not quite satisfactory, because `Rope.toTextLines` materialises too much, while we really care about the first line of `rope'` only. Do we care about illegal positions, where column counter is bigger than a corresponsing line?

michaelpj: I just pushed a version that just pulls out a single line and operates on that. I think that avoids the issue of materializing to much stuff. I'd appreciate a look.

> Do we care about illegal positions, where column counter is bigger than a corresponsing line?

We probably shouldn't crash, but it's not clear what we should do. Return `Nothing`, perhaps.

michaelpj: My current implementation will return the position of the end of the line. I'll put a note to fix that.

lsp/src/Language/LSP/VFS.hs
```diff
@@ -346,6 +352,52 @@ changeChars logger str start finish new = do
...
+      (_, suffix) = Rope.splitAtLine (fromIntegral l) text
+      (utf16Line, _) = Rope.splitAtLine 1 suffix
+
+      utfLine = URope.fromText $ Rope.toText utf16Line
+      -- Split at the given position in *code points*
+      (utfLinePrefix, _) = URope.splitAt (fromIntegral cpc) utfLine
+      utf16LinePrefix = Rope.fromText $ URope.toText utfLinePrefix
+      -- Get the length of the prefix in *code units*
+      cuc = Rope.length utf16LinePrefix
+  in J.Position (fromIntegral l) (fromIntegral cuc)
```
Bodigrim, line 380:
```suggestion
  in J.Position l (fromIntegral cuc)
```

lsp/src/Language/LSP/VFS.hs
```diff
@@ -346,6 +352,52 @@ changeChars logger str start finish new = do
...
+positionToCodePointPosition vFile (J.Position l cuc) = do
+  let text = _file_text vFile
+      (_, suffix) = Rope.splitAtLine (fromIntegral l) text
+      (utf16Line, _) = Rope.splitAtLine 1 suffix
+
+  -- Split at the given position in *code units*
+  (utf16LinePrefix, _) <- Rope.splitAt (fromIntegral cuc) utf16Line
+  let utfLinePrefix = URope.fromText $ Rope.toText utf16LinePrefix
+      -- Get the length of the prefix in *code points*
+      cpc = URope.length utfLinePrefix
```
Bodigrim, line 396: It's a bit cleaner to avoid `URope.fromText . Rope.toText` conversions:
```suggestion
  let utfLinePrefix = Rope.toTextLines utf16LinePrefix
      -- Get the length of the prefix in *code points*
      cpc = ULines.length utfLinePrefix
```

michaelpj: I guess I thought it would avoid the reliance on your unreleased change, but sure.

Bodigrim: Ah, right, makes sense. Conversion of a single line should not be expensive.

Bodigrim: If we can stick to the released API, let's not block on me making a new release of `text-rope`.

michaelpj: That's the hope. No doubt this could still be cheaper, but at this point we've bounded it to logarithmic in the number of lines and linear in the length of the line, which isn't too bad.

michaelpj:
Okay, I think I'm done with this. Perhaps @wz1000 you could take a look and approve.

michaelpj:
Thanks Pepe!

PR #408, opened 2022-02-27
Add `Semigroup`/`Monoid` instances for `LspT`

Gabriel439:


michaelpj:
A short justification would be nice :)

Do people generally like these instances? I agree they're certainly lawful, but I worry about getting worse error messages because you can now `<>` a monadic  thing when you maybe didn't mean to.

lsp/src/Language/LSP/Server/Core.hs
```diff
@@ -64,6 +65,12 @@ import           Control.Monad.Catch (MonadMask, MonadCatch, MonadThrow)
 newtype LspT config m a = LspT { unLspT :: ReaderT (LanguageContextEnv config) m a }
   deriving (Functor, Applicative, Monad, MonadCatch, MonadIO, MonadMask, MonadThrow, MonadTrans, MonadUnliftIO, MonadFix)
 
+instance Semigroup a => Semigroup (LspT config m a) where
+    (<>) = liftA2 (<>)
+
+instance Monoid a => Monoid (LspT config m a) where
```
michaelpj, line 71: Derive these `via Ap`?

Gabriel439:
I wrote up a post to explain why I typically add these instances: https://www.haskellforall.com/2022/03/applicatives-should-usually-implement.html

michaelpj:
Thanks, that was a lot more explanation than I was expecting :) LGTM

michaelpj (approved):

michaelpj:
Needs some language pragmas too, alas.

michaelpj:
Thanks!

PR #412, opened 2022-03-14
Fix missing 'namespace' token type

### LINKED ISSUE #410: missing 'namespace' type in SemanticTokenTypes and knownSemanticTokenTypes
It probably got missed because it's listed all by itself above the doc comment for the 'type' token type in the enum at https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_semanticTokens ...

michaelpj:
Fixes #410.


michaelpj:
Note that indeed this changes some test output, because the serialization of the token types with the default legend depends on ordering, so we would have been sending everything off-by-one before this :(

PR #415, opened 2022-04-03
Add hashable instance for location, range, position

OliverMadine:
Add hashable instances to remove orphaned instances from the rename-plugin in the [recent merge](https://github.com/haskell/haskell-language-server/pull/2809) to the HLS.


michaelpj:
LGTM. I guess ideally we would do these consistently, but I'm happy to just add them on an ad-hoc basis for now.

michaelpj (approved):

PR #413, opened 2022-03-16
fix typos

askeblad:
lsp-test/src/Language/LSP/Test.hs
lsp-types/src/Language/LSP/Types/Lens.hs
lsp-types/src/Language/LSP/Types/Parsing.hs
lsp-types/src/Language/LSP/Types/SMethodMap.hs
lsp/src/Language/LSP/Server/Core.hs

<hr>

for another PR this line could be rewritten

lsp-test/src/Language/LSP/Test/Replay.hs

`-- | Swaps out any commands uniqued with process IDs to match the specified process ID`

PR #416, opened 2022-04-07
fix spelling errors

askeblad:
lsp-test/src/Language/LSP/Test.hs
lsp/src/Language/LSP/Server/Core.hs
lsp/src/Language/LSP/VFS.hs
lsp/test/URIFilePathSpec.hs

@michaelpj missed these in `#413`

### DIFF
--- a/lsp-test/src/Language/LSP/Test.hs
+++ b/lsp-test/src/Language/LSP/Test.hs
@@ -241,7 +241,7 @@ runSessionWithHandles' serverProc serverIn serverOut config' caps rootDir sessio
   exitServer :: Session ()
   exitServer = request_ SShutdown Empty >> sendNotification SExit Empty
 
-  -- | Listens to the server output until the shutdown hack,
+  -- | Listens to the server output until the shutdown ACK,
   -- makes sure it matches the record and signals any semaphores
   listenServer :: Handle -> SessionContext -> IO ()
   listenServer serverOut context = do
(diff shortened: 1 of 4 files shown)


lsp-test/src/Language/LSP/Test.hs
```diff
@@ -241,7 +241,7 @@ runSessionWithHandles' serverProc serverIn serverOut config' caps rootDir sessio
   exitServer :: Session ()
   exitServer = request_ SShutdown Empty >> sendNotification SExit Empty
 
-  -- | Listens to the server output until the shutdown ack,
+  -- | Listens to the server output until the shutdown hack,
```
michaelpj, line 244: This is wrong, it was "ack" as in "acknowledges"

michaelpj (approved):

lsp-test/src/Language/LSP/Test.hs
askeblad: could it be typed as `SHUTDOWN ACK`

michaelpj: I think it's clear enough to me, but I wouldn't object to ACK.

michaelpj: You can just push more commits to your branch, and it will update the PR.

askeblad: don't you have to squash

michaelpj: No.

michaelpj:
Probably "lossily"

askeblad:
> Probably "lossily"

would spj concur

then I would submit a PR for the cabal repo with that change

askeblad:
> Probably "lossily"

thanks

opened a new cabal repo PR

Issue #414, opened 2022-03-30
Changing the configuration after `workspace/didChangeConfiguration` requests


pgujjula:
I notice that `lsp` now has a function `getConfig` to get the client configuration "as set via the `initialize` and `workspace/didChangeConfiguration` requests." I am a little confused how the configuration is supposed to be updated, however.

Our language server has an implementation of `onConfigurationChange` that parses the changed configuration. When we try changing the configuration, the `lsp` library complains with an error `lsp:no handler for:  SWorkspaceDidChangeConfiguration`. I assume this means that we should also provide an implemention for the `workspace/didChangeConfiguration` notification. However, it's unclear to me what this handler should be doing, since it seems all the relevant work is already done by the `onConfigurationChange` function.

Therefore, my question is, what functionality should the `onConfigurationChange` implementation provide, what functionality should the `workspace/didChangeConfiguration` handler provide, and what is the difference between them?

pgujjula:
I'd like to bump this question, @michaelpj perhaps?

michaelpj:
What we currently do with `onConfiguraitonChange` is:
1. If the `intialize` request contains `initializationOptions`, we call it on that (this is somewhat questionable but some clients expect it).
2. We put it in the `LanguageContextEnv`.

So at the moment you need to add your own `didChangeConfiguration` notification handler, which may well just pull `onConfigurationChange` from the `LanguageContextEnv` and call it. I _think_ the reason we don't install a handler by default is that a server may need to do other things in the `didChangeConfiguration` handler, e.g. invalidating state. So it's something we can't write for people.

pgujjula:
Thanks for the information! I'll try implementing our configuration management using this info and get back to you if I have any more questions!

pgujjula:
Ok, I've come across a problem with this setup. When the configuration is changed in VSCode, the client sends a `workspace/didChangeConfiguration` notification with a null body (see https://github.com/microsoft/vscode-languageserver-node/issues/380#issuecomment-414691493), and expects the server to re-read the configuration from the client.

When the server receives a `workspace/didChangeConfiguration` message, it seems that it calls the parser specified in `onConfigurationChange` to parse the config, updates the stored config, and also calls the user-specified handler for `SWorkspaceDidChangeConfiguration`.

[...]

PR #418, opened 2022-04-13
Add setConfig as an analogue to getConfig

pgujjula:
As per the discussion in #414, we need a `setConfig` function for the user to be able to properly manage the configuration.


lsp/src/Language/LSP/Server/Core.hs
```diff
@@ -441,12 +441,18 @@ freshLspId = do
...
 -- | The current configuration from the client as set via the @initialize@ and
--- @workspace/didChangeConfiguration@ requests.
+-- @workspace/didChangeConfiguration@ requests, as well as by calls to
+-- 'setConfig'.
 getConfig :: MonadLsp config m => m config
 getConfig = getsState resConfig
 
 {-# INLINE getConfig #-}
 
+setConfig :: MonadLsp config m => (config -> (a,config)) -> m a
```
michaelpj, line 451: This isn't really what we'd normally call `setConfig`, which would have type `config -> m ()`. I think we should either expose that function, or call this `stateConfig` or something weird like that. (the whole state manipulation dance here is a bit odd, but that's how it is...)

pgujjula: Understood, I've updated the PR accordingly.

Issue #419, opened 2022-04-13
How to improve the documentation of lsp-types


BinderDavid:
As a user of the lsp library I often have to peruse the documentation of the types defined in the lsp-types library. The current documentation at https://hackage.haskell.org/package/lsp-types-1.4.0.1/docs/Language-LSP-Types.html is quite difficult to digest, since the haddocks contain almost every type that the library exports. For example, I cannot even open the documentation page on my (quite old) tablet without crashing the browser tab.

Since I often use the library, I would volunteer some of my time to improve the documentation of the lsp-types library. But before starting with this I would like to have clarification on the best direction of travel:

Could the modules in https://github.com/haskell/lsp/blob/master/lsp-types/src/Language/LSP/Types.hs be exposed modules, with the `Language.LSP.Types` module just reexporting the other modules? This would make it much simpler to browse the haddocks, and to document each module individually. The current separation into different modules is already quite logical, since it mostly maps one LSP request to one module.

As an alternative solution, only those modules clearly corresponding to a single LSP requests could be exposed. This would makes it possible to reorganize the remaining modules in the future without having breaking changes.

michaelpj:
> Could the modules in https://github.com/haskell/lsp/blob/master/lsp-types/src/Language/LSP/Types.hs be exposed modules, with the Language.LSP.Types module just reexporting the other modules?

I think that's probably fine at this point. The organization of those modules is somewhat arbitrary, so it was nice to not expose them and not have to worry about keeping them stable and sensible, but I think the rate of change isn't so high now, so it's probably safe enough!

Ideally we would just follow the organization of the spec, but it's pretty terribly organized too :sweat_smile:

michaelpj:
This problem is only getting worse. Sadly, I don't know of an obvious way to fix it :/

Moreover, it's tricky because moving to generate everything from the metamodel (https://github.com/haskell/lsp/pull/458) pushes us to define everything in one module. At that point it would be a lot of manual re-exporting to put a pretty face on it. Not sure what to do.

PR #420, opened 2022-04-14
Extend the list of exposed modules of lsp-types by those modules whic…

BinderDavid:
The documentation of the types defined in the lsp-types library could be improved (cf. #419). As a first step, more modules should probably be exposed. As one step in this direction, this PR exposes all the modules which correspond to "Language features" on the official documentation site of the LSP protocol: https://microsoft.github.io/language-server-protocol/specification
The mapping is not 1:1, since the official documentation is not super sensibly organized either, but maybe this choice of exposed modules is a sensible first step.

### DIFF
--- a/lsp-types/lsp-types.cabal
+++ b/lsp-types/lsp-types.cabal
@@ -18,24 +18,14 @@ extra-source-files:  ChangeLog.md, README.md
 
 library
   exposed-modules:     Language.LSP.Types
+                     , Language.LSP.Types.CallHierarchy
                      , Language.LSP.Types.Capabilities
-                     , Language.LSP.Types.Lens
-                     , Language.LSP.Types.SMethodMap
-                     , Data.IxMap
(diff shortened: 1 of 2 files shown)


michaelpj:
I wonder if we should just expose them all.

Another alternative would be to move the modules to `Langage.LSP.Types.Internal....` and then expose them, so that we give the impression that they may not be stable (which is true).

BinderDavid:
>so that we give the impression that they may not be stable (which is true).
My impression was that for the modules exposed in this PR, there is the least likelihood that they would change, since they correspond very closely both to the LSP spec and only contain the definitions specific to one LSP request. (The types might evolve with the evolution of the LSP spec, of course, but is it likely that LSP requests are removed or merged?).

michaelpj:
I think this is now obsolete, but we discussed at Zurihac and I'd be open to a new iteration of this.

PR #422, opened 2022-04-18
Modularise the server input and output

michaelpj:
The goal here is to make the `Control` module as boring and dispensible
as possible, so that users can put the pieces together as they like.
Thisi s a step in that direction, tackling the server in/out threads.


michaelpj:
Huh, I think the `lsp-tests` now genuinely seem to not terminate on Windows. How baffling.

PR #424, opened 2022-04-26
Add instance: Hashable SomeLspId

thomasjm:
Just an instance I've had to include separately in projects in the past.

### DIFF
--- a/lsp-types/src/Language/LSP/Types/LspId.hs
+++ b/lsp-types/src/Language/LSP/Types/LspId.hs
@@ -7,9 +7,10 @@
 module Language.LSP.Types.LspId where
 
 import qualified Data.Aeson                                 as A
-import           Data.Text                                  (Text)
+import           Data.Hashable
 import           Data.Int (Int32)
 import           Data.IxMap
+import           Data.Text                                  (Text)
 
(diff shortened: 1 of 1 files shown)


lsp-types/src/Language/LSP/Types/LspId.hs
```diff
@@ -45,3 +46,6 @@ instance Ord SomeLspId where
       go (IdString a) (IdString b) = a `compare` b
       go (IdInt    _) (IdString _) = LT
       go (IdString _) (IdInt    _) = GT
+
+instance Hashable SomeLspId where
+  hashWithSalt n (SomeLspId a) = hashWithSalt n (toBase a)
```
michaelpj, line 51: I'm a little confused. Why not derive `Hashable` for `LspId` and then just define this in terms of that?

thomasjm: Oh sure, that works too

thomasjm: I have no idea why it is like this actually, this is old code haha. Sending that change now...

michaelpj (approved):
Lovely, thanks.

PR #425, opened 2022-04-26
Fix #268 by introducing a MaybeN type

### LINKED ISSUE #268: Spec compliance: null vs missing values
I just noticed that some types don't perfectly obey the LSP spec regarding when a value is allowed to be missing vs. when it must be present and `null`.

For example, if you look at [InitializeParams](https://microsoft.github.io/language-server-protocol/specification#initialize), the `processId` field has type `processId: integer | null`, while the `locale` field is written `locale?: string`. Thus the latter is allowed to be missing but the former is not.

All optional fields in this package seem to be represented as `Maybe`, and encoded to JSON with `omitNothingFields = True`: 

https://github.com/alanz/lsp/blob/36b655d9cc538f79b14c847ca383d97ffcc18162/lsp-types/src/Language/LSP/Types/Utils.hs#L109
https://github.com/alanz/lsp/blob/36b655d9cc538f79b14c847ca383d97ffcc18162/lsp-types/src/Language/LSP/Types/Initialize.hs#L44

This caused a problem for me when trying to write a Haskell client for a Python language server powered by [pygls](https://github.com/openlawlibrary/pygls). Upon sending an `initialize` message with an empty `processId`, I get an NPE because it assumes the value will be present:

```
    self._server.process_id = params.processId
AttributeError: 'Object' object has no attribute 'processId'
```

To be compliant with the spec, perhaps we need to replace these "required optional values" with a different `Maybe`-like wrapper that always encodes to `null` when empty?

thomasjm:
Taking a stab at fixing #268 by introducing a new `MaybeN` type, which is just like `Maybe` except that `Nothing` values get encoded to JSON `null`. This is inspired by the existing `List` type.

This can be used to conform properly to the LSP spec, which sometimes specifies types like `a | null` and sometimes `a?` (i.e. omit the value when absent).

The downside of this approach is that for every field you convert from `Maybe` to `MaybeN`, you have to go through all usages of that field and change them. I've included an example of doing this for `InitializeParams`; for full compliance we would have to do the same for all `Maybe` fields that are supposed to be `a | null`.

Of course, it's unpleasant to expose this encoding detail in the field types and make all these changes. This seemed like the most straightforward way, but a cleaner approach might be possible with TH. For example, maybe we could extend `makeExtendingDataType` so that it takes extra arguments indicating which type of `Maybe` to use, and generates suitable `FromJSON`/`ToJSON` instances manually? FWIW a search through the LSP spec indicates about 40 cases of `| null` to handle.

### DIFF
--- a/lsp-test/src/Language/LSP/Test.hs
+++ b/lsp-test/src/Language/LSP/Test.hs
@@ -200,14 +200,14 @@ runSessionWithHandles' serverProc serverIn serverOut config' caps rootDir sessio
   let initializeParams = InitializeParams Nothing
                                           -- Narrowing to Int32 here, but it's unlikely that a PID will
                                           -- be outside the range
-                                          (Just $ fromIntegral pid)
+                                          (JustN $ fromIntegral pid)
                                           (Just lspTestClientInfo)
-                                          (Just $ T.pack absRootDir)
-                                          (Just $ filePathToUri absRootDir)
+                                          (JustN $ T.pack absRootDir)
+                                          (JustN $ filePathToUri absRootDir)
                                           (lspConfig config')
                                           caps
                                           (Just TraceOff)
-                                          (List <$> initialWorkspaceFolders config)
+                                          (maybeN (List <$> initialWorkspaceFolders config))
   runSession' serverIn serverOut serverProc listenServer config caps rootDir exitServer $ do
     -- Wrap the session around initialize and shutdown calls
     initReqId <- sendRequest SInitialize initializeParams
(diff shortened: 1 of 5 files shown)


michaelpj:
I was actually thinking we could tackle this alongside https://github.com/haskell/lsp/issues/421. If we completely overhaul the type generation to go off the new `metaModel.json`, then we can also hopefully fix this issue in a systematic way.

thomasjm:
Oh wow, that would be nice!

michaelpj:
Sadly I don't know when I'm going to get to that :/ So maybe we should do this anyway.

thomasjm:
That would be awesome, feel free to hit me up if there's any way I can help.

michaelpj:
@thomasjm how big a problem are these issues for you? I'm chipping away at the metamodel generation approach, but it'll take a while.

thomasjm:
Well, every time I file one of these PRs it's because I've encountered a compliant LSP server in the wild that doesn't work with `lsp-types` :). I'm trying to work with a number of these.

It's not super urgent to merge these because I just use my own fork, but I thought I'd try to contribute them all back. I'm not sure if anyone else is experiencing LSP compliance pain. If the metamodel thing can bring compliance to 100% in one fell swoop it will be awesome for increasing confidence in this library.

michaelpj:
Should be fixed in master

PR #436, opened 2022-05-22
Migrate VFS to Data.Text.Utf16.Rope.Mixed

### LINKED ISSUE #423: Migrate VFS to Data.Text.Utf16.Rope.Mixed
I've added a new flavour of `Rope`, which offers simultaneous indexing by Unicode code points and by UTF-16 code units without conversions. @michaelpj could you please give it a try? If it works satisfactory, I'll release it as `text-rope-0.2`.

https://github.com/Bodigrim/text-rope/tree/1f721c1936cbfb1ea4a742f12754047afdcff3a8

Bodigrim:
Closes #423 
@michaelpj if this looks good, I'll proceed with `text-rope-0.2` release.

### DIFF
--- a/lsp/src/Language/LSP/VFS.hs
+++ b/lsp/src/Language/LSP/VFS.hs
@@ -81,9 +81,10 @@ import           Data.Ord
 import qualified Data.HashMap.Strict as HashMap
 import qualified Data.Map.Strict as Map
 import           Data.Maybe
-import qualified Data.Text.Rope as URope
-import           Data.Text.Utf16.Rope ( Rope )
-import qualified Data.Text.Utf16.Rope as Rope
+import           Data.Text.Lines as Char ( Position(..) )
+import           Data.Text.Utf16.Lines as Utf16 ( Position(..) )
+import           Data.Text.Utf16.Rope.Mixed ( Rope )
+import qualified Data.Text.Utf16.Rope.Mixed as Rope
 import           Data.Text.Prettyprint.Doc hiding (line)
 import qualified Language.LSP.Types           as J
 import qualified Language.LSP.Types.Lens      as J
@@ -112,7 +113,7 @@ data VFS = VFS { _vfsMap :: !(Map.Map J.NormalizedUri VirtualFile)
                } deriving Show
 
 data VfsLog =
-  SplitInsideCodePoint Rope.Position Rope
+  SplitInsideCodePoint Utf16.Position Rope
   | URINotFound J.NormalizedUri
   | Opening J.NormalizedUri
   | Closing J.NormalizedUri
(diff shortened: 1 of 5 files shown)


lsp/src/Language/LSP/VFS.hs
```diff
@@ -380,103 +381,42 @@ data CodePointRange =
...
-  -- Split at the given position in *code units*
-  (prefix, _) <- Rope.splitAt offset rope
-  -- Convert the prefix to a rope using *code points*
-  let utfPrefix = URope.fromText $ Rope.toText prefix
-      -- Get the length of the prefix in *code points*
-  pure $ URope.length utfPrefix
-
 -- | Given a virtual file, translate a 'CodePointPosition' in that file into a 'J.Position' in that file.
 --
--- Will return 'Nothing' if the requested position is out of bounds of the document.
```
michaelpj, line 437: What happens now? I'm guessing we get the "clamped" result. I think both behaviours are reasonable but I'd like it to be documented.

lsp/test/VspSpec.hs
```diff
@@ -313,12 +313,12 @@ vspSpec = do
       positionToCodePointPosition vfile (J.Position 1 2) `shouldBe` Nothing
       positionToCodePointPosition vfile (J.Position 1 3) `shouldBe` Just (CodePointPosition 1 2)
       positionToCodePointPosition vfile (J.Position 1 4) `shouldBe` Just (CodePointPosition 1 3)
-      positionToCodePointPosition vfile (J.Position 1 5) `shouldBe` Just (CodePointPosition 1 4)
+      positionToCodePointPosition vfile (J.Position 1 5) `shouldBe` Just (CodePointPosition 2 0)
```
michaelpj, line 316: This one looks like an actual logic change :thinking: What happened here? Was the old implementation wrong?

lsp/test/VspSpec.hs
```diff
@@ -328,16 +328,16 @@ vspSpec = do
...
-      codePointPositionToPosition vfile (CodePointPosition 1 0) `shouldBe` Just (J.Position 1 0)
-      codePointPositionToPosition vfile (CodePointPosition 1 1) `shouldBe` Just (J.Position 1 1)
-      codePointPositionToPosition vfile (CodePointPosition 1 2) `shouldBe` Just (J.Position 1 3)
-      codePointPositionToPosition vfile (CodePointPosition 1 3) `shouldBe` Just (J.Position 1 4)
-      codePointPositionToPosition vfile (CodePointPosition 1 4) `shouldBe` Just (J.Position 1 5)
+      codePointPositionToPosition vfile (CodePointPosition 1 0) `shouldBe` J.Position 1 0
+      codePointPositionToPosition vfile (CodePointPosition 1 1) `shouldBe` J.Position 1 1
+      codePointPositionToPosition vfile (CodePointPosition 1 2) `shouldBe` J.Position 1 3
+      codePointPositionToPosition vfile (CodePointPosition 1 3) `shouldBe` J.Position 1 4
+      codePointPositionToPosition vfile (CodePointPosition 1 4) `shouldBe` J.Position 2 0
```
michaelpj, line 335: ditto

michaelpj (approved):
Code looks fine, much simpler, thanks! Just a couple of questions.

lsp/test/VspSpec.hs
Bodigrim: I tend to think that the previous implementation was not quite correct. 

`vfile` here contains `a𐐀b\na𐐀b\n`. If you take 1 line (which is `a𐐀b\n`) and then 4 code points, you consume the last newline (as a codepoint) as well and now your position is `(2, 0)`.

lsp/src/Language/LSP/VFS.hs
Bodigrim: Yes, it's clamped.

lsp/test/VspSpec.hs
michaelpj: Okay, there's an interesting question here, namely: if you have a position `(x,y)` where `y` is greater than the number of characters in the line, does that represent a position on the next line? I would say no: the first coordinate of the position limits you to that line, and the most you can do is go to the end of it. Which I would represent as `(x, maxCol+1)`, although arguably that's a bit weird.

Bodigrim: Imagine that you pressed →  key `x` times and then → key `y` times. I believe in the majority of modern editors this moves your cursor to the next line `x+1`. Such convention also allows to make [`splitAtPosition`](https://hackage.haskell.org/package/text-rope-0.1/docs/Data-Text-Lines.html#v:splitAtPosition) total, which is a nice property.

michaelpj: So the more realistic example in our case where we can end up with positions out of range is the following:
- HLS has stale file contents where line 2 only has 30 characters
- The user has locally edited line 2 so they have more than 30 characters, and so their editor sends us positions with offsets over 30

It would be more wrong in general I believe to just start wrapping onto the next line. Keeping things strictly line-based is better here. That's why I initially was returning `Nothing` for positions beyond the end of the line!

Bodigrim: I see your point, but IMHO once VFS is out of sync all bets are off anyways. I think `positionToCodePointPosition` is too low-level to be bothered with this kind of sanity checks, it does not edit anything per se, but I can try reverting to the previous design.

michaelpj: Well, the VFS being slightly out of sync is actually the normal state when the user is editing the file!

I wish I had some clear design to point you to (the LSP spec sadly provides no guidance)... I think the clamping part is fine, but I do think we should try and keep things to the same line and use the position "just past" the end of the line when we go off the end.

Bodigrim:
Closing, I don't have interest or time to pursue this further. Feel free to reuse/rebase if needed.

PR #438, opened 2022-06-18
Add lower bounds for Diff and network-uri

Bodigrim:
Otherwise `Network.URI` and `Data.Algorithm.DiffOutput` are not available yet.


michaelpj (approved):

Bodigrim:
As a Hackage trustee, I made a matching revision to earlier releases:
https://hackage.haskell.org/package/lsp-types-1.4.0.1/revisions/

michaelpj:
Thank you!

PR #446, opened 2022-07-25
use OsPath in NormalizedFilePath

### LINKED ISSUE #445: Migrate to OsPath
As suggested by @pepeiborra, we have a better alternative to `FilePath` now. See: https://hasufell.github.io/posts/2022-06-29-fixing-haskell-filepaths.html

And changing the implementation of `NormalizedFilePath` might greatly impact HLS performance, according to the existing comment.
https://github.com/haskell/lsp/blob/a41954e356117d8f5cb64b20e1a9d32e8fc6a884/lsp-types/src/Language/LSP/Types/Uri.hs#L158-L160

I'm planning to:
1. Modify lsp-types to use `OsPath` in a fork
2. Create a hls branch using that fork
3. Profiling, to verify the performance (especially memory) impact

kokobd:
Closes #445. Current status:
- existing tests pass
- breaks backward compatibility due to the necessity to add MonadThrow to the type of some functions
- use `filepath-compat` to avoid CPP, and always use `OsPath` in `NormalizedFilePath`
- provide two sets of functions for `NormalizedFilePath`. One for `FilePath`, another for `OsPath`


lsp-types/src/Language/LSP/Types/Uri.hs
```diff
@@ -151,37 +184,78 @@ platformAdjustToUriPath systemOS srcPath
...
+#endif
+
 -- | Newtype wrapper around FilePath that always has normalized slashes.
 -- The NormalizedUri and hash of the FilePath are cached to avoided
 -- repeated normalisation when we need to compute them (which is a lot).
 --
 -- This is one of the most performance critical parts of ghcide, do not
 -- modify it without profiling.
-data NormalizedFilePath = NormalizedFilePath NormalizedUri !FilePath
+data NormalizedFilePath = NormalizedFilePath !NormalizedUri {-# UNPACK #-} !ShortByteString
```
michaelpj, line 199: How much of a difference does this make over just doing `!OsPath`? The unpack is nice, I guess the annoyance is that if we leave it in we'll get a warning in the case where it's `FilePath`? Or is `OsPath` not unpackable also? Just trying to keep this simple, if we don't do the unpacking then we save on some more CPP...

lsp-types/src/Language/LSP/Types/Uri.hs
```diff
@@ -151,37 +184,78 @@ platformAdjustToUriPath systemOS srcPath
...
+    v <- Data.Binary.get :: Get ShortByteString
+    let nuri = internalNormalizedFilePathToUri (wrapOsPath v)
+    return (NormalizedFilePath (fromJust nuri) v)
+
+-- Remove the CPP hack when OsPath gets 'Binary' instance.
+-- See: https://gitlab.haskell.org/haskell/filepath/-/issues/122
+
+unwrapOsPath :: OsPath -> ShortByteString
+#ifdef OS_PATH
+#if defined(mingw32_HOST_OS) || defined(__MINGW32__)
```
michaelpj, line 216: hmm, is `HOST_OS` the right thing here? isn't that the build machine not the target?

lsp-types/src/Language/LSP/Types/Uri.hs
```diff
@@ -151,37 +184,78 @@ platformAdjustToUriPath systemOS srcPath
...
+unwrapOsPath = getWindowsString . getOsString
+#else
+unwrapOsPath = getPosixString . getOsString
+#endif
```
michaelpj, line 220: are these exhaustive?

lsp-types/src/Language/LSP/Types/Uri.hs
kokobd: The point here is that I always want to use `ShortByteString` as the underlying type, no matter if the exposed type is `OsPath` or `FilePath`. `{-# UNPACK #-}` is only an optimization though. Then all versions of GHC benefit from the same performance gain, in terms of memory usage.

lsp-types/src/Language/LSP/Types/Uri.hs
kokobd: `#if defined(mingw32_HOST_OS) || defined(__MINGW32__)` is copied from `filepath`, so I bet it's correct.

lsp-types/src/Language/LSP/Types/Uri.hs
kokobd: I think it is. Below is how `PlatformString` defined in [`filepath`](https://hackage.haskell.org/package/filepath-1.4.100.0/docs/src/System.OsString.Internal.Types.html#OsString).
```
#if defined(mingw32_HOST_OS) || defined(__MINGW32__)
type PlatformString = WindowsString
#else
type PlatformString = PosixString
#endif
```

lsp-types/src/Language/LSP/Types/Uri.hs
michaelpj: Sorry, I'm still slightly confused. Which of these is it?
1. Using `OsPath` provides no memory usage benefits even with new `filepath`, so we need to use `ShortByteString` regardless to get benefits.
2. Using `OsPath` does provide memory usages in itself, but if we use `ShortByteString` we can get the same improvements for old GHCs as well.

I do like this a lot better than the `filepath-compat` version, so I think I'm probably okay with doing this even if it's only for 2.

lsp-types/src/Language/LSP/Types/Uri.hs
```diff
@@ -151,37 +184,75 @@ platformAdjustToUriPath systemOS srcPath
           FPP.addTrailingPathSeparator (init drv)
       | otherwise = drv
 
+#ifdef OS_PATH
```
michaelpj, line 187: I suggest isolating all the CPP and filepath related stuff in an internal `System.OsPath.Compat` module or something.

lsp-types/src/Language/LSP/Types/Uri.hs
kokobd: It's 2. Another option would be conditionally using `ShortByteString`/`OsPath` in `NormalizedFilePath`. But I think it will introduce more CPP. A stable type is easier to work with.

michaelpj: Or just only getting the performance improvements in later GHCs. But the changes here seem localized enough that it's probably worth doing.

lsp-types/src/Language/LSP/Types/Uri.hs
kokobd: What about `Language.LSP.Types.Uri.OsPathCompat`? Not choosing a general name like `System.OsPath.Compat` is because the CPP stuff introduced is probably only suitable for `NormalizedFilePath`.

Most other modules should simply switch between `FilePath` and `OsPath`. A type alias and a conditional import should suffice. Let's worry about that when ghc 9.6 is released.

The main distinction is that most modules only need to care about the "correctness" `OsPath` provides, not the performance gain.

michaelpj: Sure, although we probably don't have to worry too much about the name if we don't export the module.

lsp-types/src/Language/LSP/Types/Uri.hs
kokobd: The application is compiled on `BUILD_OS`, and runs on `HOST_OS`. So that's correct. The compiler's `TARGET_OS` is the application's `HOST_OS`. Applications don't have `TARGET_OS`.

michaelpj: Right, I always get this wrong :sweat_smile:

kokobd:
@mergify refresh

kokobd:
@Mergify refresh

kokobd:
I added a dimension to the test matrix, but Mergify is waiting for the old one. Strange.

lsp-types/src/Language/LSP/Types/Uri.hs
```diff
@@ -190,21 +196,24 @@ instance Hashable NormalizedFilePath where
...
-    fromString = toNormalizedFilePath
+filePathToNormalizedFilePath :: MonadThrow m => FilePath -> m NormalizedFilePath
+filePathToNormalizedFilePath fp = OsPath.fromFilePath fp >>= toNormalizedFilePath
+
+normalizedFilePathToFilePath :: MonadThrow m => NormalizedFilePath -> m FilePath
+normalizedFilePathToFilePath nfp = OsPath.toFilePath $ fromNormalizedFilePath nfp
 
-toNormalizedFilePath :: FilePath -> NormalizedFilePath
-toNormalizedFilePath fp = normalizedFilePath nuri nfp
+toNormalizedFilePath :: MonadThrow m => OsPathCompat -> m NormalizedFilePath
```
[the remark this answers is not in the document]
michaelpj: So the failure here ultimately comes from `decodeUtf`, right? Which ultimately represents the fact that a filepath is bytes and a URI should be text. So this is a meaningful failure. We could offer a partial version that really throws, but that also seems like a footgun for us in future.

The only other alternative I can see is to have URI/NormalizedUri also contain an `OsPath` i.e. bytes rather than text. Unclear how bad that would be.

kokobd:
@michaelpj After some rewrites, I'm finally happy with the current version. The most notable change is that now we always store a UTF-8 encoded `ShortByteString` in `NormalizedFilePath`, so that `toNormalizedFilePath :: FilePath -> NormalizedFilePath` is still total.

lsp-types/src/Language/LSP/Types/Uri.hs
```diff
@@ -191,20 +209,36 @@ instance Hashable NormalizedFilePath where
...
   where
-      nfp = FP.normalise fp
-      nuri = internalNormalizedFilePathToUri nfp
+    nfp = FP.normalise fp
+    nuri = internalNormalizedFilePathToUri nfp
 
-fromNormalizedFilePath :: NormalizedFilePath -> FilePath
-fromNormalizedFilePath (NormalizedFilePath _ fp) = fp
+-- | Extracts 'FilePath' from 'NormalizedFilePath'.
+-- The function is total. The 'HasCallStack' constraint is added for debugging purpose only.
```
michaelpj, line 222: This is a little confusing as it immediately calls error! I get the argument: the invariant of NFP is that it contains a correctly encoded string so this should never fail.

However, I also bet we call this a lot. What would happen if we also cached the decoded filepath in `NormalizedFilePath`? Would it blow out memory usage?

lsp-types/src/Language/LSP/Types/Uri/OsPath.hs
```diff
@@ -0,0 +1,47 @@
...
+osPathToNormalizedFilePath :: MonadThrow m => OsPath -> m NormalizedFilePath
+osPathToNormalizedFilePath = fmap toNormalizedFilePath . unsafePerformIO' . decodeFS
+
+{-|
+Extracts 'OsPath' from 'NormalizedFilePath'. Throws 'IOException' if the conversion fails.
+-}
+normalizedFilePathToOsPath :: MonadThrow m => NormalizedFilePath -> m OsPath
+normalizedFilePathToOsPath = unsafePerformIO' . encodeFS . fromNormalizedFilePath
+
+unsafePerformIO' :: (MonadThrow m, NFData a) => IO a -> m a
```
michaelpj, line 37: doc?

lsp-types/test/URIFilePathSpec.hs
```diff
@@ -249,7 +265,7 @@ genValidUnicodeChar = arbitraryUnicodeChar `suchThat` isCharacter
   where isCharacter x = x /= '\65534' && x /= '\65535'
 
 normalizedFilePathSpec :: Spec
-normalizedFilePathSpec = do
+normalizedFilePathSpec = beforeAll (setFileSystemEncoding utf8) $ do
```
michaelpj, line 268: Does something actually fail if you don't do this?

lsp-types/src/Language/LSP/Types/Uri/OsPath.hs
```diff
@@ -0,0 +1,47 @@
...
+import           Control.Monad.Catch
+import           Language.LSP.Types.Uri
+import           System.IO.Unsafe       (unsafePerformIO)
+import           System.OsPath
+
+{-|
+Constructs 'NormalizedFilePath' from 'OsPath'. Throws 'IOException' if the conversion fails.
+-}
+osPathToNormalizedFilePath :: MonadThrow m => OsPath -> m NormalizedFilePath
+osPathToNormalizedFilePath = fmap toNormalizedFilePath . unsafePerformIO' . decodeFS
```
michaelpj, line 29: I'm getting a little confused, would you mind writing a comment somewhere explaining the deal with the various encodings? The idea is something like: we always decode OsPaths using the system encoding into a string, and then we re-encode them as UTF8 to store them in NFP?

michaelpj (approved):
Looking plausible! I worry that maybe this extra encoding/decoding might lose some of the benchmark improvements you were getting?

michaelpj:
Oops, I missed the label! oh well, if you want to address any of the comments you can do a follow-up

.github/workflows/haskell.yml
```diff
@@ -41,6 +50,12 @@ jobs:
 
     - name: Cabal update
       run: cabal update
+    - name: Cabal configure
+      shell: bash
+      run: |
+        if [ ${{ matrix.ospath }} = "true" ]; then
```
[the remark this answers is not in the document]
kokobd: Nice idea!

lsp-types/src/Language/LSP/Types/Uri.hs
```diff
@@ -151,27 +164,32 @@ platformAdjustToUriPath systemOS srcPath
...
--- The NormalizedUri and hash of the FilePath are cached to avoided
+-- | A file path that is already normalized. It is stored as an UTF-8 encoded 'ShortByteString'
+--
+-- The 'NormalizedUri' is cached to avoided
 -- repeated normalisation when we need to compute them (which is a lot).
 --
 -- This is one of the most performance critical parts of ghcide, do not
 -- modify it without profiling.
-data NormalizedFilePath = NormalizedFilePath NormalizedUri !FilePath
+data NormalizedFilePath = NormalizedFilePath !NormalizedUri {-# UNPACK #-} !ShortByteString
```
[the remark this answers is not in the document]
kokobd: Yeah, it is mentioned in some lines above.

lsp-types/test/URIFilePathSpec.hs
kokobd: Yes, default file system encoding ignores the failure. See: https://hackage.haskell.org/package/base-4.17.0.0/docs/src/GHC.IO.Encoding.html#initFileSystemEncoding
image

So in the real world, it won't throw any exception. (But I want to test the exception)

lsp-types/src/Language/LSP/Types/Uri.hs
kokobd: Yes, memory usage is the point of using `ShortByteString` or `OsPath` instead of `FilePath`. Despite the encoding/decoding, **both** CPU time and memory usage decreased in benchmarks!

I saw most use sites of `toNormalizedFilePath` just extracts the `FilePath` and pass it to some IO function, so the CPU is not the bottleneck.

hasufell:
> Looking plausible! I worry that maybe this extra encoding/decoding might lose some of the benchmark improvements you were getting?

Does this still use old directory/base for file operations? Because all of them do encoding/decoding at the FFI boundary. The only thing you'll gain is lower memory footprint due to avoiding String.

Can't say which one has the bigger impact.

lsp-types/src/Language/LSP/Types/Uri.hs
```diff
@@ -151,27 +164,32 @@ platformAdjustToUriPath systemOS srcPath
...
+        return (NormalizedFilePath (internalNormalizedFilePathToUri v') v)
+
+encodeFilePath :: String -> ShortByteString
+encodeFilePath = BS.toShort . T.encodeUtf8 . T.pack
 
--- | A smart constructor that performs UTF-8 encoding and hash consing
-normalizedFilePath :: NormalizedUri -> FilePath -> NormalizedFilePath
-normalizedFilePath nuri nfp = NormalizedFilePath nuri nfp
+decodeFilePath :: ShortByteString -> Either UnicodeException String
+decodeFilePath = fmap T.unpack . T.decodeUtf8' . BS.fromShort
```
pepeiborra, line 192: We can always speed this up with something like `utf8-string` or rolling our own utf-8 decoding

pepeiborra: https://hackage.haskell.org/package/utf8-string-1.0.2/docs/Codec-Binary-UTF8-String.html

lsp-types/src/Language/LSP/Types/Uri/OsPath.hs
pepeiborra: I have missed all the discussion in the PR, so I need a comment explaining why we don't do the obvious thing (which would be storing the `OsPath` inside `NormalizedFilePath`).

pepeiborra, line 29: Instead of `decodeFS` you could use the pure `decodeWith` and isolate the `unsafePerformIO` in a single global:

```
osPathToNormalizedFilePath :: MonadThrow m => OsPath -> m NormalizedFilePath
 osPathToNormalizedFilePath = fmap toNormalizedFilePath . unsafePerformIO' . decodeWith systemEnc

{-# NOINLINE systemEnc #-}
systemEnc :: TextEncoding
systemEnc = unsafePerformIO getFileSystemEncoding
```

kokobd: Applied in #453. Thank you!

PR #448, opened 2022-08-14
lsp-test: Add MonadThrow instance to Session

thomasjm:
A `MonadThrow` instance is useful in the `Session` monad so that you can make assertions in various test frameworks.

This adds `exceptions` to the dependencies for `lsp-test`, but it's a pretty common dependency (already used in `lsp`).

### DIFF
--- a/lsp-test/lsp-test.cabal
+++ b/lsp-test/lsp-test.cabal
@@ -51,6 +51,7 @@ library
                      , data-default
                      , Diff >= 0.3
                      , directory
+                     , exceptions
                      , filepath
                      , Glob >= 0.9 && < 0.11
                      , lens
@@ -101,7 +102,7 @@ test-suite func-test
   main-is:             FuncTest.hs
   hs-source-dirs:      func-test
   type:                exitcode-stdio-1.0
-  build-depends:       base 
+  build-depends:       base
                      , lsp-test
                      , lsp
                      , process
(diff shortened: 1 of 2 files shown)


pepeiborra:
What kind of assertions do you have in mind? Can you share an example?

thomasjm:
Pretty much all test frameworks I'm aware of throw exceptions to indicate failures. For example, HUnit:

https://hackage.haskell.org/package/HUnit-1.6.2.0/docs/Test-HUnit-Base.html#v:assertEqual

pepeiborra:
I'm interested in understanding your use case. Would you be able to share an example?

thomasjm:
I'm testing an LSP server. Here's an example test:

```haskell
  it "does document highlight" $ doSession documentHighlightCode $ \filename -> do
    ident <- openDoc filename "haskell"
    let desired = [
          DocumentHighlight (Range (Position 0 0) (Position 0 3)) (Just HkWrite)
          , DocumentHighlight (Range (Position 1 9) (Position 1 12)) (Just HkRead)
          ]
    getHighlights ident (Position 0 1) >>= (`shouldBe` List desired)
```

The `doSession` command here runs some setup and ultimately calls `runSessionWithConfig` with the callback, so the body of this test runs in the `Session` monad. The `shouldBe` call is the part that needs `MonadThrow` to work.

pepeiborra:
We use `liftIO`, because the hunit assertion combinators that we use are not lifted to `MonadThrow`. I can see how the lifted version enabled by `MonadThrow` is nicer, so this change seems Ok

michaelpj (approved):
Especially since the whole point of this is to be used in testing, this seems like a worthwhile improvement.

michaelpj:
Oh, if you wanted to be helpful it looks like there's some stuff in the `lsp-test` tests that could be cleaned up with this!

PR #449, opened 2022-08-14
lsp-test: export runSessionWithConfig' to allow modifying the CreateProcess

thomasjm:
The default `runSessionWithConfig` launches an LSP server process, but doesn't provide any flexibility for modifying the `CreateProcess` to be used. This prevented me from doing things like tweaking the environment variables for the process.

This PR exports a new function `runSessionWithConfig'`, where the first argument allows the caller to modify the `CreateProcess`.

### DIFF
--- a/lsp-test/src/Language/LSP/Test.hs
+++ b/lsp-test/src/Language/LSP/Test.hs
@@ -25,6 +25,7 @@ module Language.LSP.Test
     Session
   , runSession
   , runSessionWithConfig
+  , runSessionWithConfig'
   , runSessionWithHandles
   , runSessionWithHandles'
   -- ** Config
@@ -133,7 +134,7 @@ import System.Environment
 import System.IO
 import System.Directory
 import System.FilePath
-import System.Process (ProcessHandle)
+import System.Process (ProcessHandle, CreateProcess)
 import qualified System.FilePath.Glob as Glob
 import Control.Monad.State (execState)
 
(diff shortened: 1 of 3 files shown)


lsp-test/src/Language/LSP/Test.hs
```diff
@@ -158,9 +159,19 @@ runSessionWithConfig :: SessionConfig -- ^ Configuration options for the session
                      -> FilePath -- ^ The filepath to the root directory for the session.
                      -> Session a -- ^ The session to run.
                      -> IO a
-runSessionWithConfig config' serverExe caps rootDir session = do
+runSessionWithConfig = runSessionWithConfig' id
+
+-- | Starts a new session with a custom configuration.
+runSessionWithConfig' :: (CreateProcess -> CreateProcess)
```
michaelpj, line 165: Can we give this a more descriptive name? `runSessionInCustomProcess` or something?

I also wonder if it would make more sense to just take a user-specified `CreateProcess` rather than taking a modification function? We could also then remove e.g. the `serverExe` argument, which is somewhat clean. I'm not sure what the nicest way to factor this is.

michaelpj (commented):
It's not clear to me what the nicest way to let people configure something deep inside this function like this. My best alternative is just letting people pass the `CreateProcess` directly.

thomasjm:
> My best alternative is just letting people pass the `CreateProcess` directly.

FWIW callers can already use a preexisting process if they call the lower-level `runSessionWithHandles`, which is similar but leaves the (somewhat tricky) handle setup to the caller.

As for the caller passing the `CreateProcess`, my thought was that that would be worse because then the caller has no way to inspect or change the extra changes that may be applied by the library (in [Language.LSP.Test.Server](https://github.com/haskell/lsp/blob/master/lsp-test/src/Language/LSP/Test/Server.hs)). Currently those changes just involve setting up the streams but that could potentially change, and it would be frustrating for the library to be changing your `CreateProcess` in mysterious ways you can't adjust. So I went for maximum generality.

lsp-test/src/Language/LSP/Test.hs
thomasjm: Sure, happy to tweak the name once we're settled on the design! Responded to the other suggestion below

michaelpj:
> Currently those changes just involve setting up the streams but that could potentially change, and it would be frustrating for the library to be changing your CreateProcess in mysterious ways you can't adjust. So I went for maximum generality.

Okay, so I guess I'm wondering why you don't just call the `withHandles` version? Yes, it's a little bit of extra code, but not so much? I doubt we're really likely to do much setup except for spawning a process and setting up streams in future, so it's not like you're at that much risk of missing out on useful things that the library does for you.

thomasjm:
> Okay, so I guess I'm wondering why you don't just call the `withHandles` version?

Well, I just have a healthy fear of setting up these streams. If you do it wrong you can get weird blocking behavior or delays in message processing. I'm not even 100% sure this library does it optimally now, since there are a few places in `haskell-language-server` where the stderr is set to `LineBuffering` rather than `NoBuffering`. 

As I use this library more I sort of anticipate tweaking `Server.hs` for this reason, and in any case I think getting it right is exactly what this library should take care of rather than burdening the user. Of course you're right that I could do it that way though.

michaelpj:
Fair enough!

michaelpj (approved):
I'm happy with this so long as we give it a slightly more descriptive name.

thomasjm:
Great! I renamed it `runSessionWithConfigCustomProcess` and fixed up the Haddocks, should be ready to merge.

PR #450, opened 2022-08-21
Expose runSessionWithHandles'

pepeiborra:
Needed for https://github.com/haskell/haskell-language-server/pull/3117


kokobd (approved):

michaelpj:
Maybe we should give these things better names, but :shrug:

PR #451, opened 2022-08-23
add post jobs to github workflows

kokobd:
Mergify relies on branch protection settings, which can only be configured here:
image

When a PR changes the test matrix, Mergify will still wait for the old jobs that no longer exist. See #446 for an example.

HLS repo [uses post jobs for each workflow and sets branch protection rules to wait for those post jobs](https://github.com/haskell/haskell-language-server/pull/3117/checks?check_run_id=7942452334).

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -49,3 +49,14 @@ jobs:
     - if: matrix.ghc == '8.10.7' && runner.os == 'Linux'
       name: Build using stack
       run: stack build
+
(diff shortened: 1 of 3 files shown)


kokobd:
@michaelpj Do you have access to the repository settings page? If so, can you set the branch protection rule of the master branch to check `haskell_post_job` and `nix_post_job` instead?

michaelpj (approved):

michaelpj:
Done!

PR #452, opened 2022-08-24
remove constraint on entropy

kokobd:
It's fixed now.


michaelpj:
Hmm, we should also pin the index state really.

michaelpj (approved):

PR #455, opened 2022-09-08
Handle custom error codes in ResponseError + support spec 3.17

thomasjm:
At present the `ErrorCode` type only handles the reserved codes defined in the spec, but the server is free to use its own codes. So I extended it with a constructor `ErrorCodeCustom Int`. Without this change, parsing error messages with non-reserved codes fails.

Also, I added `ServerCancelled` and `RequestFailed`, which were added in spec version 3.17.


michaelpj (approved):
Thanks!

PR #457, opened 2022-09-17
Add Ord instance for some types

July541:
These instances are used in hls.

https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-call-hierarchy-plugin/src/Ide/Plugin/CallHierarchy/Internal.hs#L191-L193

https://github.com/haskell/haskell-language-server/blob/master/plugins/hls-call-hierarchy-plugin/test/Main.hs#L423-L424


July541:
Please rerun the test, it failed in some strange way.

michaelpj (approved):

PR #458, opened 2022-09-19
Generate types from metamodel

michaelpj:
This generates all the LSP types from the LSP metamodel, a new machine-readable format they have produced.

This requires a pretty formidable amount of TH, but in return we get a hopefully much better maintenance process, and we won't always be behind or wrong as we have been. There are also some nice touches, like on GHC
>9.2 we can now actually attach doucmentation to things we generate with
TH, so we can transfer all the documentation from the metamodel to Haddock.

This is going to result in a _lot_ of churn, enough so that I think it should be a new top-major version of `lsp-types`. In particular:

- Various inconsistencies with the spec got fixed.
- We aim for scrupulously matching the spec, even when that is weird, so there's a lot more `A |? Null` rather than `Maybe A` than there used to be.
- The new version makes heavy use of anonymous records to handle the anonymous structure literals in the spec.
- I took a fairly heavy-handed approach to avoiding name clashes in a reasonably future-proof way, but that means that a bunch of names changed, notably the constructors of `Method`/`SMethod`.

This still needs a bunch of work:
- [ ] `lsp-test` tests are broken for some reason
- [ ] More documentation and tidying in `CodeGen`
- [ ] Changelog
- [ ] Explain how to update the metamodel in the README
- [ ] Tidy up the generated documentation so that links go to the generated types not the LSP spec.
- [ ] Ensure everything builds on all versions of GHC we support.

Plus a bunch of other TODOs in the code. Any help would be welcome, but here are the top things I'm interested in feedback on:

1. How does the new version look? Is the organization okay? I'll try and get some Haddock with the new documentation in place soonish so people can look at that without building it.
2. The gnarly TH. Anyone who wants to dive in is more than welcome, it's not _too_ bad but it would be nice if we could figure out a way to make it less gnarly.


michaelpj:
@thomasjm @drsooch might be of interest

michaelpj:
@sjakobi I think you're the main maintainer of the dhall language server, which is one of the main non-HLS users of this package that I know of. I'd really appreciate your thoughts!

michaelpj:
Guide for reviewers:
- Metamodel stuff is in the `Metamodel` hierarchy, including the types for reading the metamodel and the code generation itself, which is the most gnarly part
   -  The easiest way to review the generated code at the moment is either to build the haddock or to use `-ddump-splices` on the `Generated` module.
- Most of the rest of the `Language.LSP.Types` hierarchy is just a cut down version of what was there before due to much of it being generated
    - There is some new stuff in `Message` to handle some typed/untyped versions of things
- Everything in the rest of the project is just adapting to the changes. Might be interesting to take a glance at to see the sort of changes that this will cause.

sjakobi:
> @sjakobi I think you're the main maintainer of the dhall language server, which is one of the main non-HLS users of this package that I know of. I'd really appreciate your thoughts!

Unfortunately, I'm not very familiar with the `dhall-lsp-server` internals at all – I've mostly been bumping bounds. Maybe @Gabriella439 has thoughts on this change.

Gabriella439:
My take is that you should just do whatever you need to do.  We'll figure out a way to adapt downstream

lsp-test/src/Language/LSP/Test.hs
```diff
@@ -618,169 +625,158 @@ applyEdit doc edit = do
...
     Right x -> x
-    Left err -> throw $ UnexpectedResponseError (SomeLspId $ fromJust $ rsp ^. LSP.id) err
+    Left err -> throw $ UnexpectedResponseError (SomeLspId $ fromJust $ rsp ^. LSP.id) (toUntypedResponseError err)
 
 -- | Applies formatting to the specified document.
 formatDoc :: TextDocumentIdentifier -> FormattingOptions -> Session ()
 formatDoc doc opts = do
   let params = DocumentFormattingParams Nothing doc opts
-  edits <- getResponseResult <$> request STextDocumentFormatting params
+  edits <- absorbNull . getResponseResult <$> request SMethod_TextDocumentFormatting params
```
drsooch, line 732: I know this is lsp-test code but this line pops up a bunch

lsp-types/src/Language/LSP/MetaModel/Types.hs
```diff
@@ -0,0 +1,225 @@
...
+
+A typescript version of the metamodel types can be found here, which is useful for constructing
+this Haskell model of them:
+https://github.com/microsoft/vscode-languageserver-node/blob/main/tools/src/metaModel.ts
+-}
+module Language.LSP.MetaModel.Types where
+
+import           Data.Aeson         hiding (Null, String)
+import qualified Data.Aeson         as JSON
+import qualified Data.Aeson.TH      as JSON
```
drsooch, line 25: nit: some mixture between qualified and unqualified aeson functions in this file

lsp-types/src/Language/LSP/MetaModel/CodeGen.hs
```diff
@@ -0,0 +1,911 @@
...
+genStruct s@Structure{name, documentation, since, proposed} = do
+  let structName = name
+
+  st <- asks symbolTable
+  tn <- case Map.lookup name st of
+    Just n  -> pure n
+    Nothing -> fail $ "no name for " ++ show name
+  addDocumentation tn documentation since proposed
+
+  ctor <- lift $ makeToplevelName name
```
drsooch, line 376: can just use `tn` here? Or does it have to be a new name? or does it even matter?

drsooch: Then in `genMetaModel` no need to un-mangle names?

lsp-types/src/Language/LSP/MetaModel/CodeGen.hs
```diff
@@ -0,0 +1,911 @@
...
+
+(.=?) :: (J.KeyValue kv, J.ToJSON v) => J.Key -> Maybe v -> [kv]
+k .=? v = case v of
+  Just v' -> [k J..= v']
+  Nothing -> mempty
+
+-- | Generate a type declaration corresponding to a top-level named struct.
+genStruct :: Structure -> CodeGenM [TH.Dec]
+genStruct s@Structure{name, documentation, since, proposed} = do
+  let structName = name
```
drsooch, line 368: nit: structName is used in one place only

lsp-types/src/Language/LSP/MetaModel/CodeGen.hs
```diff
@@ -0,0 +1,911 @@
...
+          Just _ -> fail "impossible: wrong custom constructor"
+          Nothing -> pure Nothing
+    let clauses = normalClauses ++ maybeToList customClause
+        sd = TH.SigD toBaseTypeN (TH.ArrowT `TH.AppT` TH.ConT tn `TH.AppT` ty)
+        fd = TH.FunD toBaseTypeN clauses
+    pure [ sd, fd ]
+
+  -- Result type is Maybe unless X allows custom values, in which case we can always interpret
+  -- a base value as a custom member
+  -- valueToX :: <base type> -> Maybe X
```
drsooch, line 301: do we want a partial function for NonCustom enums?

lsp-types/src/Language/LSP/MetaModel/Types.hs
```diff
@@ -0,0 +1,225 @@
...
+  toJSON Both           = toJSON @String "both"
+
+instance FromJSON MessageDirection where
+  parseJSON = withText "MessageDirection" $ \case
+    "serverToClient" -> pure ServerToClient
+    "clientToServer" -> pure ClientToServer
+    "both"           -> pure Both
+    t                -> fail $ "unknown message direction " ++ show t
+
+data BaseTypeName = URI | DocumentUri | Integer | UInteger | Decimal | RegExp | String | Boolean | Null
```
drsooch, line 47: nit: URI -> Uri :) Unless there is some name clashing?

lsp-types/src/Language/LSP/MetaModel/CodeGen.hs
```diff
@@ -0,0 +1,911 @@
...
+      pure (name, (n, noBang, mty))
+
+  let datad = TH.DataD [] tn [] Nothing [TH.RecC ctor (fmap snd args)] [stockDeriving]
+
+  toJsonD <- do
+    (unzip -> (args, pairEs)) <- for props $ \Property{name, optional} -> do
+      n <- lift $ TH.newName "arg"
+      pairE <- case optional of
+        Just True -> lift [| $(TH.litE $ TH.stringL $ T.unpack name) .=? $(TH.varE n) |]
+        _ -> lift [| [$(TH.litE $ TH.stringL $ T.unpack name) J..= $(TH.varE n)] |]
```
drsooch, line 395: The spec defining "optional" to be False if it's omitted is confusing and weird... but it's in the spec so

lsp-types/src/Language/LSP/MetaModel/CodeGen.hs
```diff
@@ -0,0 +1,911 @@
...
+  pure [datad, toJsonD, fromJsonD]
+
+-- | Get the list of properties of a struct, including inherited ones.
+getStructProperties :: Structure -> CodeGenM [Property]
+getStructProperties s@Structure{name, properties, extends, mixins} = do
+  st <- asks structTable
+  let
+    extends' = fromMaybe [] extends
+    mixins' = fromMaybe [] mixins
+    supertypes = extends' ++ mixins'
```
drsooch, line 421: I wonder if it's ever possible for an `extended type` or `mixin` type to have duplicate fields? Is it worth de-duping `superProps` (aside from local name de-duping)

lsp-types/src/Language/LSP/MetaModel/CodeGen.hs
```diff
@@ -0,0 +1,911 @@
...
+
+  mdN <- lift $ TH.newName "messageDirection"
+  lift $ addHaddock mdN "Get a singleton witness for the message direction of a method."
+  mtN <- lift $ TH.newName "messageKind"
+  lift $ addHaddock mtN "Get a singleton witness for the message kind of a method."
+
+  let methodName context fullName =
+        let pieces = T.splitOn "/" fullName
+        -- TODO: don't like this
+        in makeConstrName (Just context) $ foldMap capitalize pieces
```
drsooch, line 525: I don't see anything wrong with this? Could probably just default to `Nothing` instead of `Just context` none of the context appears to disambiguous

lsp-types/src/Language/LSP/MetaModel/CodeGen.hs
drsooch: nvm it breaks TH - got it

lsp-types/src/Language/LSP/Types/Capabilities.hs
```diff
@@ -1,19 +1,24 @@
...
+import           Data.Row
+import           Language.LSP.Types.Common
+import           Language.LSP.Types.CodeAction
+import           Language.LSP.Types.SemanticTokens
+import           Language.LSP.Types.Internal.Generated
+import           Prelude                               hiding (min)
+
+{-
+TODO: this is out-of-date/needs an audit
+TODO: can we generate this? process the 'since' annotations in the metamodel?
```
drsooch, line 20: Seems like a good idea to generate this as well. I'm going cross-eyed looking at it. I can give it a shot if you want

drsooch:
I honestly, don't think there is much to change. The TH generation seems sensible, I don't think there's much to change honestly. Codegen is pretty much straightforward, no matter what language you are using. I made some general code nit picks that honestly you can just ignore. There were one or two actual suggestions that I think are worth it. Otherwise lgtm.

lsp-types/src/Language/LSP/MetaModel/Types.hs
```diff
@@ -0,0 +1,225 @@
...
+  | TupleType { tItems :: [Type] }
+  | StructureLiteralType { stlValue :: StructureLiteral }
+  | StringLiteralType { slValue :: Text }
+  | IntegerLiteralType { ilValue :: Integer }
+  | BooleanLiteralType { blValue :: Bool }
+  deriving (Show, Eq, Ord)
+
+data Request = Request
+  { method              :: Text
+  , params              :: Maybe Type -- typescript says it can be [Type], but it never is so whatever
```
drsooch, line 84: I'd argue this should match the spec, better to match the spec and reduce churn later imo

lsp-types/src/Language/LSP/MetaModel/CodeGen.hs
```diff
@@ -0,0 +1,911 @@
...
+-- | Names we can't put in Haskell code.
+reservedNames :: Set.Set T.Text
+reservedNames = Set.fromList [ "data", "type" ]
+
+-- | Sanitize a name so we can use it in Haskell.
+sanitizeName :: T.Text -> T.Text
+sanitizeName n =
+  -- Names can't start with underscores! Replace that with a 'U' for lack
+  -- of a better idea
+  let n' = if "_" `T.isPrefixOf` n then T.cons 'U' $ T.tail n else n
```
drsooch, line 123: What about `Abstract` or `Base` or `Internal` or something like that. Looks like most usages are "abstract" classes in the traditional OOP sense.

drsooch:
operators
Also not sure why the operator-like constructors appear like this on my Haddock. It seems to just be for Record Fields

July541:
Any update here? I'm going to address some lsp-test parts, some new features from 3.16 spec attracted me.

michaelpj:
I have a partially-rewritten version that generates Haskell source instead. I wasn't able to get this version to work with GHC 9, and the thought of trying to make it work with multiple GHC versions made me scared. So I decided to switch to generating source. I need to finish it, basically...

michaelpj:
Closing in favour of https://github.com/haskell/lsp/pull/478

PR #462, opened 2022-10-25
Parse header message field 'Content-Type'

### LINKED ISSUE #461: Failed to parse message header
vim9 has implemented language server protocol, but the message is like:  
```
'Content-Length: 799\r\nContent-Type: application/vim-jsonrpc; charset=utf-8\r\n\r\n{"method":"initialize","jsonrpc":"2.0","id":1,"params":{"workspaceFolders":[{"uri":"file:///home/s...
```   
and haskell language server comes out:
```
2022-09-28T19:11:28.497691Z | Info | Starting server  
2022-09-28T19:11:28.499555Z | Error | Failed to parse message header : string
```  
when i send message manually like:   
```
'Content-Length: 799\r\n\r\n{"method":"initialize","jsonrpc":"2.0","id":1,"params":{"workspaceFolders":[{"uri":"file:///home/s...
```    
haskell language server starts successfully:   
```
2022-09-28T19:46:52.562221Z | Info | Starting server  
2022-09-28T19:46:52.566620Z | Info | Started LSP server in 0.01s
```  
according to the specification of [language server protocol](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#headerPart), header field 'Content-Length' and 'Content-Type' are supported, they are terminated by '\r\n'. Could you please add support for parsing 'Content-Type' header field or just ignore it?

saitmilk:
fix #461

### DIFF
--- a/lsp/src/Language/LSP/Server/Control.hs
+++ b/lsp/src/Language/LSP/Server/Control.hs
@@ -20,6 +20,7 @@ import qualified Colog.Core as L
 import           Colog.Core (LogAction (..), WithSeverity (..), Severity (..), (<&))
 import           Control.Concurrent
 import           Control.Concurrent.STM.TChan
+import           Control.Applicative((<|>))
 import           Control.Monad
 import           Control.Monad.STM
 import           Control.Monad.IO.Class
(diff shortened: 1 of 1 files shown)


michaelpj (commented):
I wonder if this would be a bit simpler if we rely on attoparsec backtracking? Can we not do something like:
```
try contentType
len <- contentLength
try contentType
...
```
or something? i.e. just try and parse it both before and afterwards? That seems like it would be a bit easier to understand.

saitmilk:
how about this one
[version2](https://github.com/haskell/lsp/compare/master...saitmilk:lsp:fix-parsing-message-header1)

michaelpj (approved):

PR #471, opened 2023-01-26
Remove confusing haddock, escape '\' in code example

jhrcek:


lsp-types/src/Language/LSP/Types/Message.hs
```diff
@@ -371,7 +371,6 @@ data ResponseError =
 
 deriveJSON lspOptions ''ResponseError
 
--- | Either result or error must be Just.
```
jhrcek, line 374: This is misleading since the former 2 Maybe fields were replaced with Either in 
https://github.com/haskell/lsp/commit/95f2e76d1e8a35cfc6bfdf73eb511768d6cf9392#diff-6031041f17e1defe329987e180cff35227ab23a55bf8fa9415d7ebb4ab5c93d2R409

The comment seems to have returned through merge mistake in 
https://github.com/haskell/lsp/commit/9e78cce4000dde4666cb04fa1fb59a38cc3d8a2a#diff-6031041f17e1defe329987e180cff35227ab23a55bf8fa9415d7ebb4ab5c93d2R368

lsp/src/Language/LSP/Server/Core.hs
```diff
@@ -304,8 +304,8 @@ data ServerDefinition config = forall m a.
       --
       -- @
       --  ServerDefinition { ...
-      --  , doInitialize = \env _req -> pure $ Right env
-      --  , interpretHandler = \env -> Iso
+      --  , doInitialize = \\env _req -> pure $ Right env
```
jhrcek, line 307: Special characters (like `\`) need to be escaped in @ code blocks: https://haskell-haddock.readthedocs.io/en/latest/markup.html#code-blocks

lsp-types/src/Language/LSP/Types/DocumentSymbol.hs
```diff
@@ -232,7 +232,7 @@ data SymbolInformation =
     , _kind          :: SymbolKind -- ^ The kind of this symbol.
     , _tags          :: Maybe (List SymbolTag) -- ^ Tags for this symbol.
     , _deprecated    :: Maybe Bool -- ^ Indicates if this symbol is deprecated. Deprecated, use tags instead.
-    -- | The location of this symbol. The location's range is used by a tool
+    , -- | The location of this symbol. The location's range is used by a tool
```
jhrcek, line 235: Fixes misplaced comment

Screenshot from 2023-01-26 13-23-11


Mixing `-- ^` and `-- |` style comments in records leads to misplaced comments.
In these cases `-- |` is not interpreted as beginning of the next comment, but as a continuation of -- ^ comment.

This issue is relatively pervasive in this package and I can think of 2 approaches to fixing these:

1. move commas to make it clear to haddock that this comment belongs to the next field 
i.e. from
```
-- | bla
, somefield :: Int
```
to
```
, -- | bla
somefield :: Int
```

3. switch all `-- ^` comments within records to `-- |` style
This would mean more lines of code changed, but also more consistency.

Would you prefer 1 or 2?

thomasjm: I think option 1 is a bit weird, maybe we switch it to 2? I personally prefer '-- |` comments.

Or as a compromise, what about this: any comment which contains at least one `-- |` comment must contain *all* `-- |` comments. And the ones that are purely `-- ^` can be left as is.

PR #479, opened 2023-06-04
Replace some TResponseError with ResponseError

joyfulmantis:
This a surgical edit to replace some of the TResponseErrors with ResponseErrors, making it easier to port HLS to the new release. This is one of the possible permanent solutions moving forward. Other possible options are: 
- Keeping the TResponseMessage as is and creating another type in between TResponseMessage and ResponseMessage. 
- Implementing these changes at the same time as also creating a new fully typed ResponseMessage. 
- Refactoring HLS to use TResponseErrors (I feel it should be possible, but will need someone more familiar with Haskell type magic than I am).

### DIFF
--- a/lsp-types/src/Language/LSP/Protocol/Message/Types.hs
+++ b/lsp-types/src/Language/LSP/Protocol/Message/Types.hs
@@ -132,7 +132,7 @@ data TResponseMessage (m :: Method f Request) =
   TResponseMessage
     { _jsonrpc :: Text
     , _id      :: Maybe (LspId m)
-    , _result  :: Either (TResponseError m) (MessageResult m)
+    , _result  :: Either ResponseError (MessageResult m)
     } deriving stock Generic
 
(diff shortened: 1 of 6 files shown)


lsp-types/src/Language/LSP/Protocol/Message/Types.hs
```diff
@@ -132,7 +132,7 @@ data TResponseMessage (m :: Method f Request) =
   TResponseMessage
     { _jsonrpc :: Text
     , _id      :: Maybe (LspId m)
-    , _result  :: Either (TResponseError m) (MessageResult m)
+    , _result  :: Either ResponseError (MessageResult m)
```
michaelpj, line 135:
```suggestion
    -- TODO: use `TResponseError m` for the error type, this will require quite a lot of adaptation downstream
    , _result  :: Either ResponseError (MessageResult m)
```

michaelpj (approved):
Thanks, this looks fine to me!

PR #480, opened 2023-06-06
Separate lens module, Have ResponseErrors also take LSPResponseErrors, and standardize SemanticToken fields

joyfulmantis:
This pull request includes three separate changes. The first is to separate the lens module into a separate exported module. Currently, this only includes the lens for the generated modules. The second change is to allow (T)ResponseErrors to take both a ResponseError and a LSPResponseError. To do this I used custom fromJSON code that will first convert the JSON to a ResponseError and then convert any CustomResponseError results to corresponding LSPResponseError types. Finally, I standardized the fields for SemanticToken.


lsp-types/src/Language/LSP/Protocol/Message/Types.hs
```diff
@@ -54,12 +52,23 @@ deriveJSON lspOptions ''RequestMessage
...
   ResponseError
-    { _code    :: ErrorCodes
+    { _code    :: LSPErrorCodes |? ErrorCodes
     , _message :: Text
     , _xdata   :: Maybe Value
     } deriving stock (Show, Eq, Generic)
 
-deriveJSON lspOptions ''ResponseError
+deriveToJSON lspOptions ''ResponseError
+instance FromJSON ResponseError where
```
michaelpj, line 61: The knowledge of why you are doing this is about to be lost! https://www.michaelpj.com/blog/2022/04/24/on-commenting-code.html

I would be tempted to write a Note here, something like:
```
{- Note [ErrorCodes and LSPErrorCodes]

Confusingly, the metamodel defines _two_ enums for error codes. One of
these covers JSON RPC errors and one covers LSP-specific errors. We want
to accept either, mostly so we can make use of the pre-specified enum values.

However, _both_ of them are listed as accepting custom values. This means
that `LSPErrorCodes |? ErrorCodes` isn't quite right: when we parse it from
JSON, if we get an error code that isn't a known value of `LSPErrorCodes`, we
will just use the custom value constructor, without trying `ErrorCodes`.

It's hard to find any other good way of representing things properly with what
we've got, so in the end we decided to patch up the JSON parsing with a custom
instance.
-}
```

lsp-test/test/Test.hs
```diff
@@ -22,8 +22,8 @@ import           Control.Monad
 import           Control.Lens hiding (List, Iso)
 import           Language.LSP.Test
 import           Language.LSP.Protocol.Message
-import           Language.LSP.Protocol.Types hiding (message, applyEdit)
-import qualified Language.LSP.Protocol.Types as LSP
+import           Language.LSP.Protocol.Types
+import qualified Language.LSP.Protocol.Lens as J
```
michaelpj, line 26: This use of `J` is very weird. I _believe_ it dates back to when it was really just the JSON types being imported. Or at least I can't think why else it was like that. I've been trying to slowly get rid of it in favour of qualifying as LSP (although that is longer...).

joyfulmantis: Would L be better? It's nice for lens stuff to just use one character because the getters tend to be pretty short too. LSP is longer and then is also used already by all the non lens stuff when it's imported qualified.

lsp-test/lsp-test.cabal
```diff
@@ -105,6 +105,7 @@ test-suite tests
     , lsp           ^>=2.0
     , lsp-test
     , mtl           <2.4
+    , parser-combinators
```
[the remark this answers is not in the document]
joyfulmantis: I got an error from hls that the package was hidden and that I needed to depend on it. There is I believe some related import, but it was weird that it wasn't a problem before.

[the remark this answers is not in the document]
michaelpj: Weird!

lsp-test/test/Test.hs
michaelpj: Sure, L seems fine.

PR #486, opened 2023-06-09
Add helper instances and functions

joyfulmantis:
These are custom instances we use in the hls package, along with some helper functions.


lsp-types/src/Language/LSP/Protocol/Types/Common.hs
```diff
@@ -123,6 +124,16 @@ nullToMaybe :: a |? Null -> Maybe a
 nullToMaybe (InL a) = Just a
 nullToMaybe (InR _) = Nothing
 
+maybeToNull :: Maybe a -> a |? Null
+maybeToNull (Just x) = InL x
+maybeToNull Nothing  = InR Null
+
+instance Semigroup s => Semigroup (s |? Null) where
```
michaelpj, line 131:
```suggestion
-- This is equivalent to the instance for 'Maybe s'
instance Semigroup s => Semigroup (s |? Null) where
```

lsp-types/src/Language/LSP/Protocol/Types/CustomInstances.hs
```diff
@@ -0,0 +1,27 @@
...
+import           Data.Semigroup                ()
+
+instance Semigroup WorkspaceEdit where
+  (WorkspaceEdit a b c) <> (WorkspaceEdit a' b' c') = WorkspaceEdit (a <> a') (b <> b') (c <> c')
+instance Monoid WorkspaceEdit where
+  mempty = WorkspaceEdit Nothing Nothing Nothing
+
+instance Hashable Location
+instance Hashable Range
+instance Hashable Position
```
michaelpj, line 16: Hmm, this one we should maybe just derive for everything?

lsp-types/src/Language/LSP/Protocol/Types/CustomInstances.hs
```diff
@@ -0,0 +1,27 @@
...
+
+instance Default ClientCapabilities where
```
michaelpj, line 18: Suggest putting these in `Capabilities.hs`

lsp-types/src/Language/LSP/Protocol/Types/CustomInstances.hs
```diff
@@ -0,0 +1,27 @@
...
+    def = ClientCapabilities (Just def) (Just def) (Just def) (Just def) (Just def) Nothing
```
michaelpj, line 19: kinda suspicious that we don't use Nothing for the default values here?

lsp-types/src/Language/LSP/Protocol/Types/CustomInstances.hs
```diff
@@ -0,0 +1,27 @@
+{-# OPTIONS_GHC -Wno-orphans #-}
+module Language.LSP.Protocol.Types.CustomInstances where
+
+import           Language.LSP.Protocol.Internal.Types
+import           Data.Default
+import           Data.Hashable
+import           Data.Semigroup                ()
+
+instance Semigroup WorkspaceEdit where
```
michaelpj, line 9: suggest putting these in `Edit.hs`

lsp-types/src/Language/LSP/Protocol/Types/CustomInstances.hs
joyfulmantis: I guess it depends on how badly it will affect compile time. It is only used with these types in the codebase.

lsp-types/src/Language/LSP/Protocol/Types/CustomInstances.hs
joyfulmantis: Looking at 1.6, it actually is `def` (not `(Just def)`) for everything here, so this is wrong.

lsp-types/src/Language/LSP/Protocol/Types/CustomInstances.hs
joyfulmantis: The reason why I didn't is because Capabilities doesn't export any of the Capability types, those are now all in Types . The advantage then of having them here is we can export instances with types.

lsp-types/src/Language/LSP/Protocol/Types/CustomInstances.hs
joyfulmantis: If we want to export instances together with Types, than Capabilties will need to stay here (or move to Types/Capabilities.hs?!). If they do stay here, it wouldn't be immediately obvious at first where new instances should be placed, with some being with their category and some being here.

michaelpj: Hmm, okay, that does make sense. In that case I suggest renaming this to "Orphans.hs" (which is a common convention for modules of extra instances that aren't with their types), and maybe we could move some of the other orphan instances in nearby files into it too.

lsp-types/src/Language/LSP/Protocol/Types/CustomInstances.hs
michaelpj: Basically, if we're not using the generic instance it seems surprising

lsp-types/src/Language/LSP/Protocol/Types/CustomInstances.hs
michaelpj: I think the compile time is still not too bad, thanks to it being very parallel.

michaelpj (approved):

PR #487, opened 2023-06-11
Fix decoding of |?

michaelpj:
Involves a horrible hack. I don't have a better idea right now, other than trying to do something clever with generics.


michaelpj:
Let's try this for now

PR #488, opened 2023-06-11
Auto-generate Hashable instances also

michaelpj:
LSPAny was problematic, but it's also totally useless so I got rid of it anyway.


joyfulmantis (approved):
LGTM

PR #489, opened 2023-06-11
Add WatchKind helpers and fix a bug with createDoc

joyfulmantis:
WatchKind isn't very well represented, it more of a set type than anything else. This pr adds helper functions to help work with it. createDoc also uses  the new helper function to fix a bug. (Originally I had hoped this would be enough to fix one failing test, and while it didn't fix the one I was hoping it would, it fixed two others)


michaelpj (approved):
LGTM, just add some haddock!

joyfulmantis:
Okay, scratch that. With the forgotten complete set added in, the test I was debugging pasts too. That just leaves one significant test failure left on ghcide.

PR #493, opened 2023-06-16
Metamodel.json needs to be in extra-source files

michaelpj:
Otherwise the metamodel component won't build.

Will need another release, I think.

### DIFF
--- a/lsp-types/lsp-types.cabal
+++ b/lsp-types/lsp-types.cabal
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               lsp-types
-version:            2.0.0.0
+version:            2.0.0.1
 synopsis:
   Haskell library for the Microsoft Language Server Protocol, data types
 
@@ -19,6 +19,7 @@ build-type:         Simple
 extra-source-files:
   ChangeLog.md
   README.md
+  metaModel.json
 
 source-repository head
   type:     git


joyfulmantis:
Why did this problem not appear in the CI or when building it from git? Why is it a mostly stackish issue now (new-hls that depends on these packages built fine with cabal)? It would be nice if we could implement some CI check, or even some manual checking to make sure that this doesn't occur again in the future.

michaelpj:
> Why did this problem not appear in the CI or when building it from git?

Cabal builds from the source distribution produced with `cabal sdist`, not from the git checkout. I took a look but I can't see an easy way to check that the sdist builds. Possibly it is checked for package candidates and I missed it? It seems like this is a common source of broken releases...

joyfulmantis:
I guess stack [has an option](https://docs.haskellstack.org/en/stable/sdist_command/#-no-test-tarball-flag) to test the produced sdist tarball. Might be something to consider for future releases

joyfulmantis (approved):

joyfulmantis:
> I guess stack [has an option](https://docs.haskellstack.org/en/stable/sdist_command/#-no-test-tarball-flag) to test the produced sdist tarball. Might be something to consider for future releases

Or I guess not, because I just tried that flag and it didn't do anything

PR #498, opened 2023-07-02
Pass the ClientCapabilities to the static handlers in ServerDefinition

michaelpj:
See changelog.


michaelpj:
I think this is an improvement, and at least in principle lets us decide what handlers to provide depending on the client caps. Perhaps a quick sanity check from @wz1000 ?

wz1000:
can't capabilities be registered dynamically?

michaelpj:
Only _server_ capabilities can be registered dynamically. AFAICT there isn't such a mechanism for client capabilities.

PR #502, opened 2023-07-07
Fix handling of optional methods

### LINKED ISSUE #501: HLS reports `LSP: no handler for: SMethod_SetTrace`
After upgrading to HLS shipped with lsp-2.0, I often see this.

I've checked the log but haven't found any info.

image

michaelpj:
Fixes #501

Ideally needs a test, but maybe @July541 can verify that it fixes the problem?

### DIFF
--- a/lsp-types/src/Language/LSP/Protocol/Message/Method.hs
+++ b/lsp-types/src/Language/LSP/Protocol/Message/Method.hs
@@ -14,6 +14,7 @@ module Language.LSP.Protocol.Message.Method where
 
 import           Data.Aeson.Types
 import           Data.Function                      (on)
+import           Data.List                          (isPrefixOf)
 import           Data.GADT.Compare
 import           Data.Proxy
 import           Data.Type.Equality
@@ -28,6 +29,11 @@ import           Unsafe.Coerce
 ---------------
 -- SomeMethod
 ---------------
+
+-- | Is this an "optional" method which servers and clients are allowed to ignore?
+isOptionalMethod :: SomeMethod -> Bool
+-- See https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#dollarRequests
+isOptionalMethod m = "$" `isPrefixOf` someMethodToMethodString m
 
 deriving stock instance Show SomeMethod
 instance Eq SomeMethod where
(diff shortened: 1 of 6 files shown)


lsp-types/src/Language/LSP/Protocol/Message/Method.hs
```diff
@@ -29,6 +30,11 @@ import           Unsafe.Coerce
 -- SomeMethod
 ---------------
 
+-- | Is this an "optional" method which servers and clients are allowed to ignore?
+isOptionalMethod :: SomeMethod -> Bool
+-- See https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#dollarRequests
+isOptionalMethod m = "$" `isPrefixOf` someMethodToMethodString m
```
July541, line 36: Where is the `/`?

July541:
I will provide feedback after a few days of use this.

lsp-types/src/Language/LSP/Protocol/Message/Method.hs
michaelpj: You're right, the spec does say `$/`, although something that started with just a `$` would be pretty confusing.

lsp/src/Language/LSP/Server/Processing.hs
```diff
@@ -404,11 +404,8 @@ handle' logger mAction m msg = do
     -- See https://microsoft.github.io/language-server-protocol/specifications/specification-current/#-notifications-and-requests.
     reportMissingHandler :: m ()
     reportMissingHandler =
-      let optional = isOptionalNotification m
+      let optional = isOptionalMethod (SomeMethod m)
       in logger <& MissingHandler optional m `WithSeverity` if optional then Warning else Error
-    isOptionalNotification (SMethod_CustomMethod p)
-      | "$/" `T.isPrefixOf` T.pack (symbolVal p) = True
-    isOptionalNotification _  = False
```
July541, line 411: I've built HLS with this fix, everything seems fine till now. But could you please tell me what's `symbolVal` wrong here, does `symbolVal` only returns `SetTrace`?

July541 (approved):
Give 👍 since I haven't seen error notification yet!

lsp/src/Language/LSP/Server/Processing.hs
michaelpj: A custom method has a `KnownSymbol` dictionary inside it that we can use to get the method string, since we can't know it statically. For the other methods, we know it statically. `someMethodToMethodString` handles all the cases. `SetTrace` used to be a custom method (i.e. we didn't list it explicitly), but now we do (because it's in the metamodel), so this code was wrong because it only worked for custom methods.

PR #512, opened 2023-08-22
Fully support use of workspace/configuration

### LINKED ISSUE #510: Support `workspace/configuration` request for every server initilize
In the latest release of `vscode-language-server-node`, they don't send `workspace/didChangeConfiguration` more when the server is initialized, which causes https://github.com/haskell/vscode-haskell/issues/920.

Following the comment of the `SynchronizeOptions` below, they suggest using `workspace/configuration`, so I'm thinking to extend `ServerDefinition` with a new field to send `workspace/configuration` while the server is initiated.

https://github.com/haskell/lsp/blob/97c76604bd5cb77dc3c683a2c4cdee0282e3e9c1/lsp/src/Language/LSP/Server/Core.hs#L280

https://github.com/microsoft/vscode-languageserver-node/blob/f2ff7d55464a1f58f978cb6635bd8865f050553c/client/src/common/configuration.ts#L132

michaelpj:
`lsp` will now rely primarily on `workspace/configuration` to get configuration from the client. See `Note [LSP configuration]` for details.

`lsp-test` also now handles `workspace/configuration` properly.

Fixes #510 

I'm working on the HLS patch, we shouldn't merge this until that's working. But I'm fairly sure that the `lsp` and `lsp-test` changes are good, so I'm putting this up for review.


lsp/src/Language/LSP/Server/Core.hs
```diff
@@ -49,23 +49,47 @@ import           Data.Text ( Text )
...
+import qualified Language.LSP.Protocol.Lens as L
+import qualified Language.LSP.Protocol.Message as L
 import           Language.LSP.Protocol.Utils.SMethodMap (SMethodMap)
 import qualified Language.LSP.Protocol.Utils.SMethodMap as SMethodMap
 import           Language.LSP.VFS
 import           Language.LSP.Diagnostics
 import           System.Random hiding (next)
 import           Control.Monad.Trans.Identity
 import           Control.Monad.Catch (MonadMask, MonadCatch, MonadThrow)
+import Prettyprinter
```
July541, line 62: Format?

lsp/src/Language/LSP/Server/Core.hs
```diff
@@ -49,23 +49,47 @@ import           Data.Text ( Text )
...
+  -- then we're going to need 'Pretty config' constraints everywhere
+  NewConfig J.Value
+  | ConfigurationParseError J.Value T.Text
+  | ConfigurationNotSupported
+  | BadConfigurationResponse ResponseError
+  | WrongConfigSections [J.Value]
+  deriving Show
+
+instance Pretty LspCoreLog where
+  pretty (NewConfig config) = "LSP: set new config:" <+> viaShow config
```
July541, line 81:
```suggestion
  pretty (NewConfig config) = "LSP: set new config: " <+> viaShow config
```

lsp/src/Language/LSP/Server/Core.hs
```diff
@@ -49,23 +49,47 @@ import           Data.Text ( Text )
...
+  pretty (ConfigurationNotSupported) = "LSP: not requesting configuration since the client does not support workspace/configuration"
+  pretty (ConfigurationParseError settings err) =
+    vsep [
+      "LSP: configuration parse error:"
```
July541, line 85:
```suggestion
      "LSP: configuration parse error: "
```

lsp/src/Language/LSP/Server/Core.hs
```diff
@@ -49,23 +49,47 @@ import           Data.Text ( Text )
...
+      , pretty err
+      , "when parsing"
```
July541, line 87:
```suggestion
      , ", when parsing: "
```

lsp/src/Language/LSP/Server/Core.hs
```diff
@@ -49,23 +49,47 @@ import           Data.Text ( Text )
...
+      , viaShow settings
+      ]
+  pretty (BadConfigurationResponse err) = "LSP: error when requesting configuration: " <+> viaShow err
+  pretty (WrongConfigSections sections) = "LSP: expected only one configuration section, got: " <+> viaShow sections
```
July541, line 91:
```suggestion
  pretty (WrongConfigSections sections) = "LSP: expect only one configuration section, got: " <+> viaShow sections
```

Maybe?

lsp/src/Language/LSP/Server/Core.hs
```diff
@@ -710,31 +752,117 @@ flushDiagnosticsBySource maxDiagnosticCount msource = join $ stateState resDiagn
...
-    editRange :: J.TextEdit J.|? J.AnnotatedTextEdit -> J.Range
-    editRange (J.InR e) = e ^. J.range
-    editRange (J.InL e) = e ^. J.range
+-- | Given a new config object, try to update our config with it.
+tryChangeConfig :: (m ~ LspM config) => LogAction m (WithSeverity LspCoreLog) -> J.Value -> m ()
+tryChangeConfig logger newConfigObject = do
+  parseCfg <- LspT $ asks resParseConfig
+  res <- stateState resConfig $ \oldConfig -> case parseCfg oldConfig newConfigObject of
+    Left err -> (Left err, oldConfig)
+    Right !newConfig -> (Right newConfig, newConfig)
```
July541, line 794: Any reason to use BangPattern here?

lsp/src/Language/LSP/Server/Core.hs
```diff
@@ -710,31 +752,117 @@ flushDiagnosticsBySource maxDiagnosticCount msource = join $ stateState resDiagn
...
+  caps <- LspT $ asks resClientCapabilities
+  let supportsConfiguration = fromMaybe False $ caps ^? L.workspace . _Just . L.configuration . _Just
+  if supportsConfiguration
+  then do
+    section <- LspT $ asks resConfigSection
+    _ <- sendRequest SMethod_WorkspaceConfiguration (ConfigurationParams [ConfigurationItem Nothing (Just section)]) $ \case
+      Right [newConfigObject] -> tryChangeConfig logger newConfigObject
+      Right sections -> logger <& WrongConfigSections sections `WithSeverity` Error
+      Left err -> logger <& BadConfigurationResponse err `WithSeverity` Error
+    pure ()
```
July541, line 818: How about using `void` here?

lsp/src/Language/LSP/Server/Processing.hs
```diff
@@ -330,7 +336,9 @@ handle :: (m ~ LspM config) => LogAction m (WithSeverity LspProcessingLog) -> SC
 handle logger m msg =
   case m of
     SMethod_WorkspaceDidChangeWorkspaceFolders -> handle' logger (Just updateWorkspaceFolders) m msg
-    SMethod_WorkspaceDidChangeConfiguration    -> handle' logger (Just $ handleConfigChange logger) m msg
+    SMethod_WorkspaceDidChangeConfiguration    -> handle' logger (Just $ handleDidChangeConfiguration logger) m msg
+    -- See Note [LSP configuration]
+    SMethod_Initialized                        -> handle' logger (Just $ \_ -> requestConfigUpdate (cmap (fmap LspCore) logger)) m msg
```
July541, line 341: We are requesting config update for every time initialized, considering downstream users, should we mark this as optional?

lsp/src/Language/LSP/Server/Processing.hs
```diff
@@ -427,17 +435,21 @@ shutdownRequestHandler :: Handler IO Method_Shutdown
...
+lookForConfigSection section (Object o) | Just s' <- o ^. at (fromString $ T.unpack section) = s'
+lookForConfigSection _ o = o
+
+-- | Handle a workspace/didChangeConfiguration request.
+handleDidChangeConfiguration :: (m ~ LspM config) => LogAction m (WithSeverity LspProcessingLog) -> TMessage Method_WorkspaceDidChangeConfiguration -> m ()
+handleDidChangeConfiguration logger req = do
+  section <- LspT $ asks resConfigSection
+  tryChangeConfig (cmap (fmap LspCore) logger) (lookForConfigSection section $ req ^. L.params . L.settings)
+  -- See Note [LSP configuration]
+  requestConfigUpdate (cmap (fmap LspCore) logger)
```
July541, line 452: I'm confused about why also we need `requestConfigUpdate` if `tryChangeConfig` succeeds.

lsp/src/Language/LSP/Server/Processing.hs
```diff
@@ -52,23 +54,24 @@ import           Data.Maybe
 import qualified Data.Map.Strict as Map
 import           Data.Text.Prettyprint.Doc
 import           System.Exit
-import           GHC.TypeLits (symbolVal)
 import           Control.Monad.State
 import           Control.Monad.Writer.Strict 
 import           Data.Foldable (traverse_)
+import Data.String (fromString)
```
July541, line 60: Format?

lsp/src/Language/LSP/Server/Core.hs
```diff
@@ -274,20 +300,36 @@ instance E.Exception ProgressCancelledException
...
     { defaultConfig :: config
       -- ^ The default value we initialize the config variable to.
-    , onConfigurationChange :: config -> J.Value -> Either T.Text config
-      -- ^ @onConfigurationChange oldConfig newConfig@ is called whenever the
-      -- clients sends a message with a changed client configuration. This
-      -- callback should return either the parsed configuration data or an error
-      -- indicating what went wrong. The parsed configuration object will be
-      -- stored internally and can be accessed via 'config'.
-      -- It is also called on the `initializationOptions` field of the InitializeParams
+    , configSection :: T.Text
```
July541, line 311: Is this always have the same value with `resConfigSection`?

lsp/src/Language/LSP/Server/Processing.hs
```diff
@@ -52,23 +54,24 @@ import           Data.Maybe
...
   | Exiting
 
 deriving instance Show LspProcessingLog
 
 instance Pretty LspProcessingLog where
   pretty (VfsLog l) = pretty l
+  pretty (LspCore l) = pretty l
   pretty (MessageProcessingError bs err) =
     vsep [
       "LSP: incoming message parse error:"
```
July541, line 77: Emmm, shall we unify by adding the extra space here?

July541 (approved):
Thanks! This pr overall is masterful, every part looks pretty well!

Here are just some of my curious questions.

lsp/src/Language/LSP/Server/Core.hs
michaelpj: `vsep` inserts spaces between the items if they're formatted on one line

lsp/src/Language/LSP/Server/Core.hs
michaelpj: Don't think so - it was there before. I'll drop it.

lsp/src/Language/LSP/Server/Processing.hs
michaelpj: `initialized` gets sent only once, which is when client receives our response to `initialize`. So this is a one-off thing.

lsp/src/Language/LSP/Server/Processing.hs
michaelpj: Hmm, that's a good question. I guess if we actually manage to parse out what's in `didChangeConfiguration` then we probably don't need to request new config :thinking: 

Or maybe we should just fully commit to using `workspace/configuration` and not look in the body of `didChangeConfiguration` at all. That seems to be the way that the spec maintainers want us to go...

lsp/src/Language/LSP/Server/Core.hs
michaelpj: Yeah, we copy over a bunch of fields from `ServerDefinition` into `LanguageContextEnv`...

lsp/src/Language/LSP/Server/Processing.hs
July541: I'm still confused here.. `tryChangeConfig` has updated the config, is there any situation we still need `requestConfigUpdate`?

michaelpj: I'll write it out in detail in the doc. Basically, all of the following are "allowable" scenarios:
1. Client supports `workspace/configuration`, sends an empty update in `didChangeConfiguration`. This is what new vscode will do, I believe. Then the first attempt will fail (nothing in `didChange`), and the second will succeed (requesting config).
2. Client does not support `workspace/configuration`, and sends updated config in `didChangeConfiguration`. This is rare, I think, but we should support it. Then the first attempt will succeed and the second attempt will fail (or not even try).

We do get some redundancy in the slightly odd case of a client that supports both ways, but I'm not too worried about that tbh.

PR #515, opened 2023-08-26
inlay-hint-support

TimWhiting:
Hi, I was playing around with this, and started working on inlay hints, but then noticed the inlay hint support was missing.

Here are the necessary changes. I only work with static handlers, so I haven't tested it with dynamic registration, but it works with the lsp I'm working with.


michaelpj (approved):
Thanks! I always forget about these capability bits. We should fill in the other ones too!

PR #523, opened 2023-10-08
Remove pointless IO from VFS

michaelpj:
Setting up the VFS required IO just so it could capture a reference to a temporary directory that could be used if we wanted to write files out. This is weird and non-compositional: we can just ask the caller where they want the files, and they can make the temporary directory if they want.

### DIFF
--- a/lsp-test/src/Language/LSP/Test/Session.hs
+++ b/lsp-test/src/Language/LSP/Test/Session.hs
@@ -282,8 +282,8 @@ runSession' serverIn serverOut mServerProc serverHandler config caps rootDir exi
   mainThreadId <- myThreadId
 
   let context = SessionContext serverIn absRootDir messageChan timeoutIdVar reqMap initRsp config caps
-      initState vfs = SessionState 0 vfs mempty False Nothing mempty (lspConfig config) mempty (ignoreLogNotifications config) (ignoreConfigurationRequests config)
-      runSession' ses = initVFS $ \vfs -> runSessionMonad context (initState vfs) ses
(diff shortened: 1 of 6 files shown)


fendor (approved):
LGTM, nice improvement!

neduard:
Nice cleanup!  Though consider in the future being a bit more careful with functions that are exposed in the API:

* Consider adding a message in the Changelog for non-backward-compatible changes and include the affected function/constructor names: in this case `initVFS` was removed and replaced with `emptyVFS`.  This saves maintainers having to `git bisect` to figure out why software stopped compiling with a new version of stack/cabal : a simple github search on `initVFS` and they would see the removal - note github search doesn't actually search through PR diffs unfortunately, so this removal doesn't automatically show up.
* A better way to have done this would have been eg: keep `initVFS` but add a new `emptyVFS` and change the code (internals) to use it.  This keeps the exposed interface backward compatible.  After one or more releases then you can remove `initVFS`.

In any case, no worries - glad `git bisect` worked!

michaelpj:
Hi, that's fair. I should mark this more explicitly but: the `lsp` packages are definitely still in an "unstable" phase of development. I have quite a lot of work that I still want to do on them, which probably means a lot more major version bumps. I realise that's not ideal, sorry.

neduard:
No worries at all!  I think as long as people can easily find when there's breaking changes they can always set pin their version and update whenever they wish.  Here I think it was a bit difficult to easily find why compilation broke (though I'm also Haskell beginner so maybe for others it would've been more intuitive)

PR #533, opened 2023-11-26
Add bounds to all dependencies

michaelpj:
We really should have these...

I'm going to make sure I can still build HLS on all versions with these in place before committing.


michaelpj:
Okay, it worked, so we can do this safely.

PR #535, opened 2023-12-01
More cabal file tidying

michaelpj:
Remove most constraints from non-library components. I'm unsure if this is quite right: possibly we should have bounds for any public component, but this is significantly more work to maintain (witness the odd bound on `sorted-list` from a test suite). I think this strikes a reasonable balance for now.

### DIFF
--- a/lsp-types/lsp-types.cabal
+++ b/lsp-types/lsp-types.cabal
@@ -551,7 +551,7 @@ executable generator
   main-is:            Main.hs
   other-modules:      CodeGen
   build-depends:
-    , base                 >=4.11 && <5
+    , base
     , containers
     , directory
     , filepath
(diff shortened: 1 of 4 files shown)


lsp-types/lsp-types.cabal
```diff
@@ -94,11 +94,14 @@ library
...
 
+  -- This version of filepath comes with GHC 9.6, so
+  -- we should be able to drop this flag once we only
+  -- support 9.6 or higher
   if flag(force-ospath)
     build-depends: filepath ^>=1.4.100.0
 
   else
-    build-depends: filepath
+    build-depends: filepath ^>=1.4
```
andreasabel, line 104: I think `filepath-1.5` is already out.

michaelpj: Yes, let's see if we can allow that

PR #539, opened 2023-12-29
Various fixes to progress handling

michaelpj:
1. Server-initiated progress should wait for the client to acknowledge

This is an old bug. Per the spec, we're not allowed to send any reports using the token if the client doesn't respond with a non-error response to our creation of the token.

This is a bit subtle, because it means we may need to delay the sending of the "begin" notification until we have received the token from the client.

2. No easy way to use client-initiated progress

This is simpler and faster than server-initiated progress since you don't need the extra message round-trip. You just need to pull out the progress token (if there is one) from the request and use that.

I did two things to make this better:
- The progress functions now take the client token if there is one. If there isn't one they still fall back to server-initiated progress.
- The server capabilities can now advertise that client-initiated progress is supported.

### DIFF
--- a/lsp/src/Language/LSP/Server/Core.hs
+++ b/lsp/src/Language/LSP/Server/Core.hs
@@ -648,7 +648,7 @@ withProgressBase indefinite title clientToken cancellable f = do
   progressState <- liftIO $ newMVar ProgressInitial
 
   -- Until we start the progress reporting, track the current latest progress in an MVar, so when
-  -- we do start we can start at the right point. 
+  -- we do start we can start at the right point.
   let initialPercentage = if indefinite then Nothing else Just 0
   initialProgress <- liftIO $ newMVar (ProgressAmount initialPercentage Nothing)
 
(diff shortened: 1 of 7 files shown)


michaelpj:
I'm actually unsure if we want to encourage people to use client-initiated progress, since I'm unsure if it allows cancelling properly: https://github.com/microsoft/language-server-protocol/issues/1873

michaelpj:
Okay, I'm going to make it off by default since it's not clear how this will all work.

lsp/src/Language/LSP/Server/Core.hs
```diff
@@ -625,97 +630,162 @@ getNewProgressId = do
...
+  Text ->
+  Maybe ProgressToken ->
+  ProgressCancellable ->
+  ((ProgressAmount -> m ()) -> m a) ->
+  m a
+withProgressBase indefinite title clientToken cancellable f = do
+  progressState <- liftIO $ newMVar ProgressInitial
+
+  -- Until we start the progress reporting, track the current latest progress in an MVar, so when
+  -- we do start we can start at the right point. This means that although we "lose
```
joyfulmantis, line 651: This comment seems to have been truncated.

joyfulmantis (approved):
LGTM

michaelpj:
I added some more tests that actually exercise a bunch of the cases, so we can be a bit more confident that this works, hopefully.

PR #542, opened 2024-01-01
Update Data.Text.Utf16.Rope  to Data.Text.Utf16.Rope.Mixed

soulomoon:
It should simplify a bit of the conversion logic. Although the asymptotic complexity doe not change.
This is the second attempt of https://github.com/haskell/lsp/pull/436

### DIFF
--- a/lsp/src/Language/LSP/VFS.hs
+++ b/lsp/src/Language/LSP/VFS.hs
@@ -413,7 +413,7 @@ line. Which is okay-ish, so long as we don't have very long lines.
 extractLine :: Rope.Rope -> Word -> Maybe Rope.Rope
 extractLine rope l = do
   -- Check for the line being out of bounds
-  let lastLine = Rope.lengthInLines rope
+  let lastLine = Utf16.posLine $ Rope.utf16LengthAsPosition rope
   guard $ l <= lastLine
   let (_, suffix) = Rope.splitAtLine l rope
       (prefix, _) = Rope.splitAtLine 1 suffix
--- a/lsp/src/Language/LSP/VFS.hs
+++ b/lsp/src/Language/LSP/VFS.hs
@@ -457,6 +457,7 @@ codePointRangeToRange vFile (CodePointRange b e) =
 -}
 positionToCodePointPosition :: VirtualFile -> J.Position -> Maybe CodePointPosition
 positionToCodePointPosition vFile (J.Position l c) = do
+  -- See Note [Converting between code points and code units]
   let text = _file_text vFile
   lineRope <- extractLine text $ fromIntegral l
   kLine <- case compare c (fromIntegral $ Rope.utf16Length lineRope) of
--- a/lsp/src/Language/LSP/VFS.hs
+++ b/lsp/src/Language/LSP/VFS.hs
@@ -405,6 +405,12 @@ linear time in the length of the (single) line.
 
 So the overall process is logarithmic in the number of lines, and linear in the length of the specific
 line. Which is okay-ish, so long as we don't have very long lines.
+
+We are not able to use the `Rope.splitAtPosition`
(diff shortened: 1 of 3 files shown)


soulomoon (commented):
—

lsp/test/VspSpec.hs
```diff
@@ -252,7 +252,7 @@ vspSpec = do
       positionToCodePointPosition vfile (J.Position 1 2) `shouldBe` Nothing
       positionToCodePointPosition vfile (J.Position 1 3) `shouldBe` Just (CodePointPosition 1 2)
       positionToCodePointPosition vfile (J.Position 1 4) `shouldBe` Just (CodePointPosition 1 3)
-      positionToCodePointPosition vfile (J.Position 1 5) `shouldBe` Just (CodePointPosition 1 4)
```
soulomoon, line 255: (1:5) is the new line.

soulomoon: Is it nessary to consider the new line in this case.

soulomoon:
I finish the migration, @michaelpj @Bodigrim .
See if this look good?

lsp/src/Language/LSP/VFS.hs
```diff
@@ -458,15 +427,12 @@ codeUnitOffsetToCodePointOffset rope offset = do
...
   -- See Note [Converting between code points and code units]
   let text = _file_text vFile
-  utf16Line <- extractLine text (fromIntegral l)
-  -- Convert the line a rope using *code points*
-  let utfLine = URope.fromText $ Rope.toText utf16Line
-
-  cuc <- codePointOffsetToCodeUnitOffset utfLine (fromIntegral cpc)
-  pure $ J.Position l (fromIntegral cuc)
+  lineRope <- extractLine text $ fromIntegral l
+  guard $ c <= fromIntegral (Rope.charLength lineRope)
```
michaelpj, line 434: I think we should include a comment about why we're not using `Rope.splitAtPosition`. The difference is quite subtle!

lsp/src/Language/LSP/VFS.hs
```diff
@@ -458,15 +427,12 @@ codeUnitOffsetToCodePointOffset rope offset = do
  the position.
```
michaelpj, line 427: Hmm, is this definitely still the same? For some reason I thought it might be faster :thinking: But maybe it's just that the code is simpler...

lsp/src/Language/LSP/VFS.hs
```diff
@@ -487,13 +453,12 @@ codePointRangeToRange vFile (CodePointRange b e) =
...
```
michaelpj, line 453: ditto

michaelpj (approved):
Great, thank you!

lsp/src/Language/LSP/VFS.hs
soulomoon: It just remove the fromText and toText part, saving the conversion, other logic still the same.
since the mixed version have two indices, one for codePoint, one for utf16

lsp/src/Language/LSP/VFS.hs
soulomoon: splitAtPosition just wrapped over the line. when column index out of range or when the column indexing at the newline char :(

soulomoon: The first out of range problem is easy to deal with,
when the column indexing at the newline char and still wrapped over the line, it becomes the pain.

soulomoon: The first out of range problem is easy to deal with,
But the second case when the column indexing at the newline char and still wrapped over the line, it becomes the pain.

soulomoon: comment added

PR #543, opened 2024-01-13
Handle clients sending "null" instead of a missing field

michaelpj:
This is a relatively common client error, so it's nice of us to follow the principle of robustness and accept it instead of failing.

It's a little bit complicated because in some case the optional field also has null as a valid value, so we have to make sure  we parse that correctly. I added some more tests so I'm pretty confident it's okay, though.


lsp-types/src/Language/LSP/Protocol/Types/Common.hs
```diff
@@ -187,3 +190,24 @@ instance Semigroup s => Semigroup (s |? Null) where
 k .=? v = case v of
   Just v' -> [fromString k J..= v']
   Nothing -> mempty
+
+{- |
+Parse a value optionally, with lenient but careful handling of 'Null':
```
joyfulmantis, line 195: Is it possible to make this comment slightly clearer? Something like 
```
If 'null' can be converted to the Null type from the JSON package then the result is 'Just JSON.Null'
If 'null' cannot be converted to the Null type from the JSON package then the result is 'Nothing'
```
maybe?

joyfulmantis, line 195: Is it possible to make this comment slightly clearer? Something like
```
If 'null' can be converted to the Null type from the JSON package then the result is 'Just JSON.Null'
If 'null' cannot be converted to the Null type from the JSON package then the result is 'Nothing'
```
maybe?

joyfulmantis (approved):
LGTM, although maybe consider rewording the comment for `.:|?` ?

lsp-types/src/Language/LSP/Protocol/Types/Common.hs
michaelpj: okay, I think I made it clearer!

PR #544, opened 2024-01-13
Generate Arbitrary instance for all types

### LINKED ISSUE #347: Arbitrary instances for types
It'd be pretty useful for LSP bindings authors (e.g.: [me](https://github.com/rhodiumdb/rhodium/blob/master/src/lsp.hpp)) to have Arbitrary instances in this library, since it'd make it easy to test serialization/deserialization roundtrip between these bindings and the bindings being developed.

michaelpj:
Really not that hard in the end, just some plumbing and fussing with the code generation. They're now exported via an additional public library `lsp-types-quickcheck`.

I also moved some of the mechanical stuff for generating extra instances just based on the type names (lenses, and now Arbitrary instances) to use TH, because that is better than the literal code generation approach when it's not horribly complicated.

I'm not yet making much use of this. The next step would be to e.g. generate tests for every type that check round-trips, but I think this is enough for now. There's also a bit of work to do to get arbitrary instances for all messages, but I'm going to do that later also.

Fixes #347


michaelpj:
Hmm, something is still screwy with the instances for message types. I'll put them back as they were for now until I can tackle them all properly.

michaelpj:
Possibly it's not really worth other people looking at this, it's mostly just codegen wrangling. The instances are all just generic-derived or newtype-derived or derived via enums.

PR #546, opened 2024-01-22
Avoid couple partial functions in lsp-test

jhrcek:
Was anyone able to figure out why there are sometimes timeout messages of the following king in hls test suite?

```
Test suite tests: RUNNING...
eval
  Property checking:                                            FAIL (66.02s)
    Timed out waiting to receive a message from the server.
    Last message received:
    {
        "jsonrpc": "2.0",
        "method": "$/progress",
        "params": {
            "token": "28",
            "value": {
                "kind": "end",
                "message": "Finished indexing 5 files"
            }
        }
    }
    Use -p '/Property checking/' to rerun this test only.
```

U briefly checked how lsp-test works and was surprised by the amount of partial functions used.
Here's a few easy to replace occurrences of fromJust + couple hard-to-resist hlint suggestions.

### DIFF
--- a/lsp-test/src/Language/LSP/Test.hs
+++ b/lsp-test/src/Language/LSP/Test.hs
@@ -421,7 +421,7 @@ sendRequest method params = do
   reqMap <- requestMap <$> ask
   liftIO $
     modifyMVar_ reqMap $
-      \r -> return $ fromJust $ updateRequestMap r id method
+      \r -> return $ fromMaybe r $ updateRequestMap r id method
 
   ~() <- case splitClientMethod method of
     IsClientReq -> sendMessage mess
--- a/lsp-test/src/Language/LSP/Test/Session.hs
+++ b/lsp-test/src/Language/LSP/Test/Session.hs
@@ -187,9 +187,9 @@ data SessionState = SessionState
   , vfs :: !VFS
   , curDiagnostics :: !(Map.Map NormalizedUri [Diagnostic])
   , overridingTimeout :: !Bool
+  , lastReceivedMessage :: !(Maybe FromServerMessage)
   -- ^ The last received message from the server.
   -- Used for providing exception information
-  , lastReceivedMessage :: !(Maybe FromServerMessage)
   , curDynCaps :: !(Map.Map T.Text SomeRegistration)
   -- ^ The capabilities that the server has dynamically registered with us so
   -- far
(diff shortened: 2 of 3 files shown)


lsp-test/src/Language/LSP/Test/Session.hs
```diff
@@ -513,7 +513,7 @@ logMsg t msg = do
   shouldColor <- asks $ logColor . config
   liftIO $ when shouldLog $ do
     when shouldColor $ setSGR [SetColor Foreground Dull color]
-    putStrLn $ arrow ++ showPretty msg
+    B.putStrLn $ arrow <> encodePretty msg
```
jhrcek, line 516: It seems pointless to convert pretty-printed json bytestring to String just to print it.
Wouldn't directly printing ByteString by faster?

lsp-test/src/Language/LSP/Test.hs
```diff
@@ -421,7 +421,7 @@ sendRequest method params = do
   reqMap <- requestMap <$> ask
   liftIO $
     modifyMVar_ reqMap $
-      \r -> return $ fromJust $ updateRequestMap r id method
+      \r -> return $ fromMaybe r $ updateRequestMap r id method
```
jhrcek, line 424: updateRequestMap returns Nothing when request with given LspId is already it.
I don't see how it makes sense to crash in such cases as opposed to returning the origial map unmodified. Similarly in the file below.

lsp-test/src/Language/LSP/Test/Session.hs
michaelpj: I have no idea, but seems sensible

lsp-test/src/Language/LSP/Test.hs
michaelpj: It's quite bad: that means we've somehow reused an ID which means stuff will break. So I think this probably _should_ throw, but maybe not by using `fromJust`

jhrcek: Alright I'll play around with it locally to see if I can reproduce the  me play around with how this works to see if I can reproduce the `  Timed out waiting to receive a message from the server.` locally and then get back to this or close this.

jhrcek:
Found [this](https://github.com/haskell/haskell-language-server/issues/3736) which answers my questions about test flakiness so closing this PR.

PR #548, opened 2024-01-28
Always try to register for didChangeConfiguration 

michaelpj:
Still needs some lsp-test work, but I think this should work.

### DIFF
--- a/lsp-test/ChangeLog.md
+++ b/lsp-test/ChangeLog.md
@@ -1,5 +1,14 @@
 # Revision history for lsp-test
 
+## Unreleased
+
+- `ignoreRegistrationRequests` option to ignore `client/registerCapability` requests, on
+  by default.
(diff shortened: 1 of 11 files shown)


michaelpj:
Okay, this is ready to go now, I think.

- `lsp` always registers for `didChangeConfiguration`
- `lsp-test` adapts to cope with the increased `registerCapability` spam
- `lsp-test` acts as a more unhelpful client and doesn't send notifications unless we ask for them

michaelpj:
Going to test this with HLS now.

michaelpj:
Got things working fine on HLS it seems, just need to get this green.

michaelpj:
BTW the explanation for the new behaviour is here: https://github.com/haskell/lsp/pull/548/files#diff-7ddecffae1af581bca5ca41258e9f1c668ae8cf3c80f1610cf6f5c34a9004e6eR538

PR #552, opened 2024-02-05
Remove getCompletionPrefix

michaelpj:
This has been in ghcide for a while, it's pointless having it here.


neduard:
Hello!  For the benefit of anyone searching, `PosPrefixInfo` was also removed as part of this commit

(mentioning this as both functions were part of the public API until `2.3.0.0` and people might be surprised when their packages stop compiling)

fwcd:
Makes sense as to why these were removed, but I admit that they were very convenient to have for Curry (which uses almost exactly the same syntax as Haskell), so I'm probably going to vendor them.

PR #566, opened 2024-04-07
lsp-test: add setIgnoringProgressNotifications

jhrcek:


michaelpj:
Did this achieve what you wanted in the tests?

jhrcek:
Not sure yet. 
I have a PR in hls repo with upgrade of hls to master version of lsp: https://github.com/haskell/haskell-language-server/pull/4166
I have a PR in my fork which uses this PR branch (based on master): https://github.com/jhrcek/haskell-language-server/pull/8

Both have test failures, but I didn't have time yet to look into those..

michaelpj:
It would be nice to know since I'll probably release the packages soon so that HLS can adapt to the changes you helpfully fixed :)

jhrcek:
@michaelpj Give me one more day. I'll try to fix up test failures in https://github.com/jhrcek/haskell-language-server/pull/9, which uses lsp patch applied before metamodel update to minimize the amount of breakage that other recent lsp changes seems to have introduced.

michaelpj:
To be clear, when I say "soon" I don't mean _that_ soon. Maybe next week or something.

jhrcek:
Status update: we decided not to merge it yet, because it would require enabling progress notifications in many hls tests.

Here's a test [PR](https://github.com/jhrcek/haskell-language-server/pull/9) that makes the hls tests pass by enabling listening to progress notifications within "waitForProgress" helper functions.
But this strategy feels too brittle (e.g. it probably allows for races where notification is sent sooner than waitForProgress is called). Probably more robust approach would require enabling progress notifications in session config of all tests that actually need to wait for progress to be completed OR refactoring some of those tests not to rely on progress.

PR #567, opened 2024-04-08
Stop doing action for handles and ignore responses after shutdown

soulomoon:
https://github.com/haskell/haskell-language-server/pull/4153

### DIFF
--- a/lsp/src/Language/LSP/Server/Processing.hs
+++ b/lsp/src/Language/LSP/Server/Processing.hs
@@ -457,7 +457,7 @@ handle' logger mAction m msg = do
   -- The reason that we do not include 'shutdown' itself here is because
   -- by the time we get the first 'shutdown' message, isShuttingDown will
   -- still be false, so we would still be able to process it.
-  -- This ensures we won't process the second 'shutdown' message and only 
+  -- This ensures we won't process the second 'shutdown' message and only
   -- process 'exit' during shutdown.
   let allowedMethod m = case (splitClientMethod m, m) of
         (IsClientNot, SMethod_Exit) -> True


lsp/src/Language/LSP/Server/Processing.hs
```diff
@@ -449,31 +451,31 @@ handle' ::
...
 handle' logger mAction m msg = do
-  maybe (return ()) (\f -> f msg) mAction
+  shutdown <- isShuttingDown
+  let allowedMethod m = case (splitClientMethod m, m) of
+        (IsClientNot, SMethod_Exit) -> True
+        (IsClientReq, SMethod_Shutdown) -> True
+        _ -> False
+
+  logger <& LspMessage (show m) `WithSeverity` Debug
+  when (not shutdown || allowedMethod m) $ maybe (return ()) (\f -> f msg) mAction
```
michaelpj, line 461: I think this would be stronger than it needs to be. In particular, it's not a problem if we e.g. update the VFS. It's really only the config callback that's potentially problematic.

It's also not enough, since updating the configuration is complicated. This handles the `didChangeConfiguration` notification, but not the response to `workspace/configuration`. So if we wanted to do this, I think we should just stick the check in `tryUpdateConfig` or something.

And I still think that servers should just make their config callbacks safe :)

soulomoon: You are right , it is not enough to stop the callback.
But I do not see any reason to do the `action before noti/req` even it is safe to do so, since the server is marked as shutdown anyway(Waitting the cleanup and exit notification).🤔

lsp/src/Language/LSP/Server/Core.hs
```diff
@@ -878,17 +878,19 @@ reverseSortEdit (L.WorkspaceEdit cs dcs anns) = L.WorkspaceEdit cs' dcs' anns
...
-    Left err -> (Left err, oldConfig)
-    Right newConfig -> (Right newConfig, newConfig)
-  case res of
-    Left err -> do
-      logger <& ConfigurationParseError newConfigObject err `WithSeverity` Warning
-    Right newConfig -> do
-      logger <& NewConfig newConfigObject `WithSeverity` Debug
-      cb <- LspT $ asks resOnConfigChange
-      liftIO $ cb newConfig
+  shutDown <- isShuttingDown
```
soulomoon, line 881: So we are stopping the callback here?

lsp/src/Language/LSP/Server/Processing.hs
soulomoon: Is this correct ?

There is push model and pull model(new one) in LSP,
The pull model:
1. `didChangeConfiguration ` sent to server 
2. `workspace/configuration` sent to client
3.  Response of configuration sent to server, we update the config.

What we are doing here is kind of mixed:
1. `didChangeConfiguration ` sent to server, we `tryChangeConfig`.
2. `workspace/configuration` sent to client
3.  Response of configuration sent to server, `tryChangeConfig` again.

soulomoon:
I think it is pretty counter intuitive that we are reponding with a invalid request but still do the actions.

michaelpj:
> I think it is pretty counter intuitive that we are reponding with a invalid request but still do the actions.

That's fair. What I am planning to do at some point is to entirely move all this stuff to be just default handlers for these methods. Then they would be treated uniformly.

I guess also that not accepting responses is pretty reasonable. We might well expect that clients won't _send_ responses if they've send `shutdown`, servers should probably just be robust to that. But we should document it in the note.

lsp/src/Language/LSP/Server/Core.hs
michaelpj: You convinced me, let's not do this and instead skip both responses and actions like you suggest.

lsp/src/Language/LSP/Server/Processing.hs
```diff
@@ -449,31 +451,30 @@ handle' ::
...
   m ()
 handle' logger mAction m msg = do
-  maybe (return ()) (\f -> f msg) mAction
+  shutdown <- isShuttingDown
+  let allowedMethod m = case (splitClientMethod m, m) of
+        (IsClientNot, SMethod_Exit) -> True
+        (IsClientReq, SMethod_Shutdown) -> True
+        _ -> False
+
+  when (not shutdown || allowedMethod m) $ maybe (return ()) (\f -> f msg) mAction
```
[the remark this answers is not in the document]
michaelpj: I think we don't need to add to the note to explain this, since we already say we ignore notifications, so it's reasonable to infer that we don't update our state or anything.

lsp/src/Language/LSP/Server/Processing.hs
```diff
@@ -449,31 +451,32 @@ handle' ::
   TClientMessage meth ->
   m ()
 handle' logger mAction m msg = do
-  maybe (return ()) (\f -> f msg) mAction
+  shutdown <- isShuttingDown
+  let allowedMethod m = case (splitClientMethod m, m) of
```
michaelpj, line 455: since this is no longer local to where it's used, let's explain
```suggestion
  -- These are the methods that we are allowed to process during shutdown.
  -- The reason that we include 'shutdown' itself here is because we change 
  -- state to "shutting down" just _before_ we process the message itself!
  let allowedMethod m = case (splitClientMethod m, m) of
```

lsp/src/Language/LSP/Server/Processing.hs
```diff
@@ -100,8 +101,9 @@ processMessage logger jsonStr = do
       FromClientMess m mess ->
         pure $ handle logger m mess
       FromClientRsp (P.Pair (ServerResponseCallback f) (Const !newMap)) res -> do
-        writeTVar pendingResponsesVar newMap
-        pure $ liftIO $ f (res ^. L.result)
+        unless shutdown <$> do
```
michaelpj, line 104:
```suggestion
        -- See Note [Shutdown]
        unless shutdown <$> do
```
and update the note to talk about responses

soulomoon: I have made a flaws here, I think we should still update the pendingResponsesVar or we get into infinite loop proccessing the same response. Ajusted and add the doc.

soulomoon:
> What I am planning to do at some point is to entirely move all this stuff to be just default handlers for these methods. Then they would be treated uniformly.
Agree, this should make things better.

lsp/src/Language/LSP/Server/Processing.hs
michaelpj: Why will we loop? AFAICT the only cost is that we will keep around response handler for _this_ response. We won't look at it again unless we get another response for that message, which we shouldn't do.

That said, it does seem neater to clean up the response handler and just not run it.

michaelpj (approved):

michaelpj:
Need to fix formatting

lsp/src/Language/LSP/Server/Processing.hs
```diff
@@ -449,31 +452,35 @@ handle' ::
...
   m ()
 handle' logger mAction m msg = do
-  maybe (return ()) (\f -> f msg) mAction
+  shutdown <- isShuttingDown
+  -- These are the methods that we are allowed to process during shutdown.
+  -- The reason that we include 'shutdown' itself here is because we change 
+  -- state to "shutting down" just _before_ we process the message itself!
+  let allowedMethod m = case (splitClientMethod m, m) of
+        (IsClientNot, SMethod_Exit) -> True
+        -- we do not allow SMethod_Shutdown to be called during shutdown
```
soulomoon, line 461: I think we should mark SMethod_Shutdown as false, so it won't be called twice.

michaelpj: That won't work, because we toggle `isShuttingDown` before we process the handler! So then we won't run the user's shutdown handler...

soulomoon: Yes, done

michaelpj (approved):
Great, thanks for making this better!

PR #568, opened 2024-04-14
Fix ghc and hlint warnings

jhrcek:


lsp-test/src/Language/LSP/Test/Files.hs
```diff
@@ -83,18 +83,13 @@ mapUris f event =
         swapDocumentChangeUri (InR (InR (InL renameFile))) = InR $ InR $ InL $ L.newUri .~ f (renameFile ^. L.newUri) $ renameFile
         swapDocumentChangeUri (InR (InR (InR deleteFile))) = InR $ InR $ InR $ swapUri id deleteFile
      in e
-          & L.changes . _Just %~ swapKeys f
+          & L.changes . _Just %~ M.mapKeys f
           & L.documentChanges . _Just . traversed %~ swapDocumentChangeUri
 
-  swapKeys :: (Uri -> Uri) -> M.Map Uri b -> M.Map Uri b
```
jhrcek, line 89: The only non-mechanical change, I just noticed and couldn't resist the simplifcation :smiley:

cabal.project
```diff
@@ -10,12 +10,15 @@ test-show-details: direct
 
 benchmarks: True
 
+package *
+  ghc-options: -Wunused-packages
```
michaelpj, line 14: Hmm. I think we normally put warning options in the cabal files, can we do that here too? Putting them all in `cabal.project` wouldn't be totally unreasonable, it's just inconsistent.

michaelpj (approved):

cabal.project
jhrcek: I was just lazy :smile: 
But I agree and I moved it to cabal files + slightly consolidating using `common warnings` stanza and fixed few more warnings.
I hope there's nothing controversial, but will wait few days for your feedback (or let you merge if you think it's ok).

lsp-types/lsp-types.cabal
```diff
@@ -53,7 +57,6 @@ library
     GADTs
     NegativeLiterals
     OverloadedStrings
-    StandaloneKindSignatures
```
jhrcek, line 56: Removing this, as it's contained in GHC2021

michaelpj:
Great, thanks! Just needs the formatting fixed.

jhrcek:
Fixed. Also noticed that I don't have right to merge (and I don't necessarily want them) so leaving merging up to you.

PR #571, opened 2024-05-06
Rewrite progress handling to allow for debouncing messages

michaelpj:
This had to be redone in order to allow us to "wake up" and notice that there are pending messages. I also wrote it so there can be a stateful interface (the `ProgressTracker`) which I think might make it easier to use in that weird case in `ghcide`. I haven't exposed that yet, though.

### DIFF
--- a/lsp/src/Language/LSP/Server/Core.hs
+++ b/lsp/src/Language/LSP/Server/Core.hs
@@ -711,6 +711,13 @@ withProgressBase indefinite title clientToken cancellable f = do
     -- for the token var to be filled if we want to use it.
     createToken :: m ()
     createToken = do
+      -- See Note [Delayed progress reporting]
+      -- This delays the creation of the token as well as the 'begin' message. Creating
+      -- the token shouldn't result in any visible action on the client side since
+      -- the title/initial percentage aren't given until the 'begin' mesage. However,
+      -- it's neater not to create tokens that we won't use, and clients may find it
+      -- easier to clean them up if they receive begin/end reports for them.
+      liftIO $ threadDelay startDelay
       case clientToken of
         -- See Note [Client- versus server-initiated progress]
         -- Client-initiated progress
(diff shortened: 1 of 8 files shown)


michaelpj:
@soulomoon what do you think? I think the `ProgressTracker` interface might be usable for that weird case in `ghcide`, since it lets you store it and call the update function from various places. The cost is that you have to deal with cancellation etc. yourself, but that seems maybe inevitable.

michaelpj:
I also need to actually test this with a live HLS and see if I can actually notice the difference!

lsp/src/Language/LSP/Server/Core.hs
```diff
@@ -138,6 +140,10 @@ data LanguageContextEnv config = LanguageContextEnv
     resState :: !(LanguageContextState config)
   , resClientCapabilities :: !L.ClientCapabilities
   , resRootPath :: !(Maybe FilePath)
+  , resProgressStartDelay :: Int
```
soulomoon, line 143: Maybe this can be set for each progress instead of globally? Since we might want different delay for different progress

soulomoon (commented):
Sounds good, it seems it is exactly the kind of flexibility HLS reports need. 
But I am not sure about the global delay.

soulomoon:
> I also need to actually test this with a live HLS and see if I can actually notice the difference!

The previous version did not work well. 
You can pick up my branch and adapt it abit to see if the new api works ? https://github.com/haskell/haskell-language-server/pull/4206

lsp/src/Language/LSP/Server/Core.hs
michaelpj: I can't actually think of a use case for this. I think it _should_ be pretty generic?

michaelpj: I'm inclined to leave it as global for now unless we have a need for it to be different.

michaelpj:
Turns out the stateful interface wasn't really that helpful and complicated things, I think https://github.com/haskell/haskell-language-server/pull/4218 is actually perfectly fine with the normal interface, arguably even clearer.

PR #573, opened 2024-05-16
fix _rootPath in initializeParams

soulomoon:


soulomoon:
It seems we have mixed `_locale` and `_rootPath`

michaelpj (approved):
Thanks! If you feel like it, you could convert this to use record construction syntax so we don't make this mistake again!

soulomoon:
Yep, that is better

michaelpj:
Needs formatting

PR #575, opened 2024-05-16
add inlay hints lsp-test support

jetjinser:
new:
- `getInlayHints`
- `getAndResolveInlayHints`
- `resolveInlayHint` (helper)

### DIFF
--- a/lsp-test/src/Language/LSP/Test.hs
+++ b/lsp-test/src/Language/LSP/Test.hs
@@ -124,6 +124,11 @@ module Language.LSP.Test (
   getAndResolveCodeLenses,
   resolveCodeLens,
 
+  -- ** Inlay Hints
+  getInlayHints,
+  getAndResolveInlayHints,
+  resolveInlayHint,
+
   -- ** Call hierarchy
   prepareCallHierarchy,
   incomingCalls,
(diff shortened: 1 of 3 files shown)


lsp-test/src/Language/LSP/Test.hs
```diff
@@ -981,6 +986,28 @@ resolveCodeLens cl = do
...
+  rsp <- request SMethod_TextDocumentInlayHint (InlayHintParams Nothing tId range)
+  pure $ absorbNull $ getResponseResult rsp
+
+{- | Returns the inlay hints in the specified range, resolving any with
+ a non empty _data_ field.
+-}
+getAndResolveInlayHints :: TextDocumentIdentifier -> Range -> Session [InlayHint]
+getAndResolveInlayHints tId range = do
+  inlayHints <- getInlayHints tId range
+  for inlayHints $ \inlayHint -> if isJust (inlayHint ^. L.data_) then resolveInlayHint inlayHint else pure inlayHint
```
michaelpj, line 1001: Hmm, this should probably actually check the server capability for resolving inlay hints, same where we do this for other resolution things. I can't remember whether it's easy to get the server capabilities in `lsp-test`, so okay to leave it like this for now unless you feel like fixing it. I'll make an issue.

michaelpj (approved):

PR #576, opened 2024-05-16
Remove dependency on uuid

SquidDev:
While the language server specification uses a UUID for dynamic registration of capabilities, it only has to be a unique id. This change removes the UUID dependency, replacing it with a home-grown random id generator.

The main motivation here is I'm trying to compile LSP (or rather, a program using it) to WASM. The UUID library pulls in a lot of superfluous dependencies (e.g. network-info and entropy) that don't compile with the wasm32-wasm version of GHC, hence trying to remove them.

### DIFF
--- a/lsp/lsp.cabal
+++ b/lsp/lsp.cabal
@@ -80,7 +80,6 @@ library
     , unliftio              ^>=0.2
     , unliftio-core         ^>=0.2
     , unordered-containers  ^>=0.2
-    , uuid                  >=1.3
 
 executable lsp-demo-reactor-server
   import:             warnings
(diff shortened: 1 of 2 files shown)


michaelpj:
This seems like a good idea. I think we could in fact do this even more simply. It just has to be a unique ID, so it can just be an incrementing integer! We in fact already have an id-counter that we use for request IDs, I think it sounds like we could just pull an id from that for registrations as well.

michaelpj:
I did that here: https://github.com/haskell/lsp/pull/578

SquidDev:
Thank you!

soulomoon:
great it might help hls to use multi-repl which need cable 3.12

PR #578, opened 2024-05-17
Use our normal auto-incrementing ids instead of UUIDs

michaelpj:
The spec just says they have to be unique IDs, not UUIDs, and we already have a tool for generating such things.


SquidDev (approved):
This is much nicer, thank you! Can we drop the `random` dependency too now?

soulomoon (approved):
LGTM!

michaelpj:
Yes, we can also remove `random`. great.

PR #587, opened 2024-06-05
Use typed response errors

### LINKED ISSUE #586: Unstructured response errors from `doInitialize`
As of lsp-2.6.0.0/lsp-types-2.2.0.0, the `doInitialize` field of `ServerDefinition` is a function that evaluates to an `IO (Either ResponseError a)`. If I'm not mistaken, the `ResponseError` component of that type could/should be a `TResponseError 'Method_Initialize`. 

This would only change the type of the `_xdata` portion of the error, refining it from its current type of `Aeson.Value` down to `InitializeError`, which would be in line with the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeResult).

michaelpj:
I have a branch adapting to this in HLS, it wasn't so bad.

Fixes #586


michaelpj:
Alright, I think this should be uncontroversial tbh.

PR #590, opened 2024-06-08
Update example code in README.md

### LINKED ISSUE #494: Bad example code in README
Specifically, this line:

```
import Language.LSP.Types
```

leads to annoying Cabal issues offering to add `lsp-types` as a dependency even when it's already there. Example code has

```
import Language.LSP.Protocol.Types
```

instead, I assume it should be updated in the README as well

patrickaldis:
Resolves #494 

The lsp protocol types changed as of [this](https://github.com/haskell/lsp/commit/4a39185ce1e49535c5050d78a328655e3cf33b02) - example docs end up being pretty far off.

Copied `lsp/example/Simple.hs` over to the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,50 +30,61 @@ typescript definitions laid out in the specification
 There are two example language servers in the `lsp/example/` folder. `Simple.hs` provides a minimal example:
 
 ```haskell
+{-# LANGUAGE DuplicateRecordFields #-}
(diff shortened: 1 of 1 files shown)


michaelpj:
Thanks! Perhaps it's just a bad idea to have literal code copied into the README, we could just refer people to the (checked) example file. Anyway, this looks good.

PR #595, opened 2024-12-09
Allow text-rope-0.3

Bodigrim:
This is starting to be an issue for Stackage.


michaelpj (approved):
Thanks!

michaelpj:
I'm not sure I want to do a release now, so I think I'll just make a Hackage revision since this is clearly okay.

michaelpj:
Revised

PR #597, opened 2024-12-23
Bump more upper bounds

Bodigrim:
Continuation of #595, apparently there are more things to bump to restore `lsp` in Stackage.


michaelpj (approved):
This is a bit more substantial, so possibly I should just do some point releases.

PR #600, opened 2024-12-31
Update CI

michaelpj:
Add 9.12, drop 9.2, enable sempahore if available.


michaelpj:
Hmm, looks like lots of stuff is blocked on moving to 9.12.

PR #605, opened 2025-02-16
Treat `undefined` and `null` initialization options the same way

### LINKED ISSUE #604: Parsing `"initializationOptions": null` should not omit a warning
When starting `haskell-language-server` using the following initialization request
```
{
...
  "initializationOptions": null,
...
}
```
I am getting the following warning, (which comes from this project, I believe):
```
2025-02-15T14:32:03.978082Z | Warning | LSP: configuration parse error:
parsing settings failed, expected Object, but encountered Null
when parsing
null
```

According to the LSP specifications, the payload should be fine though:
```
    /**
     * User provided initialization options.
     */
    initializationOptions?: LSPAny;
```

(`LSPAny` can be `null`).

Probably related: issue #543

dschrempf:
Fixes #604 .

Also:
- Overhaul `nix` configuration (add Flakes and `nix-shell` compatibility layer).
- Fix a typo.

EDIT: Seems to work on my side, has the same effect as https://github.com/haskell/haskell-language-server/pull/4498 for HLS.


michaelpj (approved):
LGTM, thanks!

PR #610, opened 2025-06-08
Track the languageId in Virtual File

VeryMilkyJoe:
The languageKind is received when a TextDocument is opened, we add this information to the VirtualFile to keep track of it.

Note that the `_language_id` is defined as a Maybe to be compatible with lsp-test's use of the virtual file system.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ghc-version: ['9.10', '9.8', '9.6', '9.4', '9.2']
+        ghc-version: ['9.12', '9.10', '9.8', '9.6', '9.4']
         # Unlikely that we'll succeed on windows and fail on macos,
         # including it is just burning CI time. But windows could have
         # path or IO issues, so worth including
--- a/cabal.project
+++ b/cabal.project
@@ -1,4 +1,4 @@
-index-state: 2024-12-23T14:00:00Z
+index-state: 2025-06-07T00:00:00Z
 
 packages:
          ./lsp
@@ -13,7 +13,5 @@ benchmarks: True
 package lsp
   flags: +demo
 
-package lsp-types
-  -- This makes a big difference here as lsp-types
-  -- has very many independent modules
-  ghc-options: -j4
+-- Useful if available
+semaphore: true
--- a/flake.lock
+++ b/flake.lock
@@ -34,11 +34,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1739446958,
-        "narHash": "sha256-+/bYK3DbPxMIvSL4zArkMX0LQvS7rzBKXnDXLfKyRVc=",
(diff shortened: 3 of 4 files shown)


lsp/src/Language/LSP/VFS.hs
```diff
@@ -92,6 +92,11 @@ data VirtualFile = VirtualFile
   -- remains in the map.
   , _file_text :: !Rope
   -- ^ The full contents of the document
+  , _language_id :: !(Maybe J.LanguageKind)
+  -- ^ The text document's language identifier
+  -- This is a Maybe, since lsp-test shares the VFS with lsp
+  -- and the server sends WorkspaceEdit notifications without a language kind.
+  -- In practical use, this should always be Just.
```
michaelpj, line 99:
```suggestion
  -- This is a Maybe, since when we use the VFS as a client
  -- we don't have this information, since server sends WorkspaceEdit 
  -- notifications without a language kind.
  -- When using the VFS in a server, this should always be Just.
```

lsp/src/Language/LSP/VFS.hs
```diff
@@ -142,8 +147,8 @@ emptyVFS = VFS mempty
 -- | Applies the changes from a 'J.DidOpenTextDocument' to the 'VFS'
 openVFS :: (MonadState VFS m) => LogAction m (WithSeverity VfsLog) -> J.TMessage 'J.Method_TextDocumentDidOpen -> m ()
 openVFS logger msg = do
-  let J.TextDocumentItem (J.toNormalizedUri -> uri) _ version text = msg ^. J.params . J.textDocument
-      vfile = VirtualFile version 0 (Rope.fromText text)
+  let J.TextDocumentItem (J.toNormalizedUri -> uri) _languageId version text = msg ^. J.params . J.textDocument
```
michaelpj, line 150:
```suggestion
  let J.TextDocumentItem (J.toNormalizedUri -> uri) languageId version text = msg ^. J.params . J.textDocument
```
The underscore prefix generally means "unused"!

lsp/src/Language/LSP/VFS.hs
```diff
@@ -158,9 +163,9 @@ changeFromClientVFS logger msg = do
     J.VersionedTextDocumentIdentifier (J.toNormalizedUri -> uri) version = vid
   vfs <- get
   case vfs ^. vfsMap . at uri of
```
michaelpj, line 165: ditto

lsp/src/Language/LSP/VFS.hs
```diff
@@ -92,6 +92,11 @@ data VirtualFile = VirtualFile
   -- remains in the map.
   , _file_text :: !Rope
   -- ^ The full contents of the document
+  , _language_id :: !(Maybe J.LanguageKind)
```
michaelpj, line 95: You might also want an accessor function for this. It's a bit random but we have them for the text and version so let's do it for this too.

VeryMilkyJoe: sounds good!

PR #617, opened 2025-07-29
[feat] allow flushing diagnostics by source and uri

### LINKED ISSUE #615: intuition for `flushDiagnosticsBySource`
Hi. I wonder three things: 
- when is it ever useful to flush everything *by source*, disregarding the `uri`? I feel like this would really hurt incrementality. If I wanted to opt in to this, then at least the second stage of the map, that contain keys by uri should be part of the `source`, right? so that I can just flush by source whenever I don't specify an uri? otherwise, if I do want to flush source and uri, I need to key by source and uri and then the next stage would only be a singleton map. 
- why is the source used as an internal criterium to store the diagnostics? Shouldn't this be unrelated to what is eventually displayed by the user? e.g. if I have several *actual* sources of a single diagnostic, then I want to still summarize it to the user as "this comes from the typechecker". It seems counterproductive to me to either having to maintain my own state around diagnostics and/or structure my program to have only one place of where diagnostics are emitted. 
- are the means to flush a single diagnostic just "not yet" implemented or are they intentionally missing? 

After reading the code I also don't quite understand if the current API is ever useful, if not intuitively, at least performance wise. Most of the operations require multiple traversals. 

To summarize, I think I would expected the following API:

`flushDiagsByDiagnosticsKey :: diagKey -> m ()`
and then 
`Map diagKey [Diagnostic]`

MangoIV:
Resolves #615

### DIFF
--- a/lsp/src/Language/LSP/Diagnostics.hs
+++ b/lsp/src/Language/LSP/Diagnostics.hs
@@ -11,6 +11,7 @@ module Language.LSP.Diagnostics (
   StoreItem (..),
   partitionBySource,
   flushBySource,
+  flushBySourceAndUri,
   updateDiagnostics,
   getDiagnosticParamsFor,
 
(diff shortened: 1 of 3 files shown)


fendor (approved):

fendor:
I got you :)

PR #635, opened 2026-03-22
Fix semantic tokens dynamic registration and CantRegister log

ijatinydv:
Fixes two small but blocking issues in the dynamic registration path.

**1. `Capabilities.hs` — semantic tokens cases were commented out**

The three semantic tokens methods fell through to `_ -> Just False` in `dynamicRegistrationSupported`, so `trySendRegistration` always returned `Nothing` for them and logged `CantRegister`. The comment said they were "messed up" due to registering with a non-existent method, but that only affects the pseudo method `textDocument/semanticTokens`. The three actual dispatch methods (`SemanticTokensFull`, `SemanticTokensFullDelta`, `SemanticTokensRange`) all map to `SemanticTokensClientCapabilities` which has `_dynamicRegistration`, so the existing `dyn m` helper works fine.

**2. `Core.hs` L623 — wrong method in CantRegister log**

`trySendRegistration` was logging `SMethod_WorkspaceDidChangeConfiguration` hardcoded instead of the actual `method` parameter. One-word fix.

These two together are the prerequisite for haskell/haskell-language-server#4084.


fendor (approved):
LGTM, thank you!

PR #640, opened 2026-05-15
Only advertise `semanticTokens` if there's a handler

9999years:
Neovim 0.12 and above will make `semanticTokens` requests on every cursor movement. Currently, the `lsp` library advertises `semanticTokens` capabilities, even if the LSP server implemented with this library doesn't a handler. This means that every time I move my cursor I get a blocking error in my editor, making it essentially unusable!

Let's only advertise the capability if there's a handler registered.

Here's a workaround in my Neovim config:

```
vim.lsp.config("hls", {
  cmd = { "static-ls" },

  on_init = function(client)
    -- static-ls (via the `lsp` Haskell library) falsely advertises
    -- semanticTokensProvider but registers no handler, producing a
    -- "no handler for: textDocument/semanticTokens/full" message on
    -- every cursor move. Strip it.
    client.server_capabilities.semanticTokensProvider = nil
  end,
})
```


fendor (approved):
LGTM, thank you!

PR #643, opened 2026-07-16
Make WebSocket support conditional

### LINKED ISSUE #642: [Feature request] Make networking support optional
When trying to port Agda Language Server to WebAssembly, a barrier I met was fighting with dependencies to compile this library. This was (and still is) by GHC WASM backend's API limitation: In [WASI](https://github.com/WebAssembly/WASI) Preview 1, the standard way of establishing a duplex connection with the host is through preopened file descriptors, either via standard I/O or via BSD-style file descriptors. It was impossible to properly implement libc's network API on top of WASI's syscalls. An early workaround to make the library compile is to [stub out all related functions](https://github.com/haskell-wasm/network/commit/1dc870889eee4ac733335ced4e274b4dfe8ed369), and is still the approach ALS takes to date.

Now that the library has been refactored to support WebSocket connections, do you consider adding flags for ~~TCP socket and~~ WebSocket support, respectively, so that library consumers can opt-out from network-related transitive dependencies? I can open a PR for review if it seems fit.

**edit**: Actually the library does not depend on `network` directly, just transitively via `websockets`. So it only needs one flag.

andy0130tw:
Fixes #642.

Makes WebSocket support configurable via a `websocket` flag (defaults to `True`). Setting this to `False` strips out the dependency of package [`websockets`](https://hackage.haskell.org/package/websockets), along with all WebSocket-related implementation, making it compatible with platforms that don't support WebSockets (e.g., the WebAssembly backend).

It introduces CPP macros to the project, which makes the code a little bit tangled. Feel free to edit it further.


fendor (approved):
LGTM!

fendor:
cc @MangoIV

andy0130tw:
The failing formatting check is due to fourmolu not knowing how to parse CPP directives? I find [this](https://github.com/tweag/ormolu#limitations) on ormolu's README:

> CPP support is experimental. CPP is virtually impossible to handle correctly, so we process them as a sort of unchangeable snippets. This works only in simple cases when CPP conditionals surround top-level declarations. See the [CPP](https://github.com/tweag/ormolu/blob/master/DESIGN.md#cpp) section in the design notes for a discussion of the dangers.

I can try to accommodate it by moving the directive around.

fendor:
@andy0130tw It is fine, the formatter check is not blocking.

michaelpj:
I would be tempted to try and avoid the CPP entirely. AFAICT the websocket support is entirely additional code. So what we can do is to a) put it in its own module, b) in the _cabal file_, expose that module only if the flag is set. I think that does everything you want without any CPP?

andy0130tw:
@michaelpj That requires a structural change, but I would be fine with that.

MangoIV (approved):
Looks like a nice and simple change.

MangoIV:
I am not a huge opponent of CPP if used sparingly.

MangoIV:
Since the websocket specific code is entirely self contained, I think it's not a detriment to the code quality at all.

fendor:
@MangoIV I also think CPP isn't a huge deal, but I agree that avoiding it altogether and having the WS parts separate could make sense any way. So I am lighlty in favour of the separate module.

@andy0130tw Did we exhaust you already or would you attempt moving the WS parts into a separate module?

andy0130tw:
@fendor I'm leaning towards *avoiding* CPP here, but I would like to passing on this task as I don't think I have enough context on the overall project architecture.

PR #644, opened 2026-07-28
Extract Websocket module to avoid CPP

fendor:
Extract optional module that exposes an API for use with websockets. Only exported if the flag `websocket` is enabled.

Follow up PR for #643 so we can discuss this further.


fendor:
The biggest difference is that we need to wrap the logger again, but I think that's fine.

fendor:
@michaelpj and @MangoIV, should we merge this? Or just use the CPP approach?

MangoIV:
@fendor thanks for doing this. I guess the only thing that I'm worried about is that this is a breaking change. If you don't mind that, I think this is good to go.

fendor:
I dont mind that :upside_down_face:

MangoIV (approved):
yeah looks relatively uncontroversial

